PR #4, opened 2017-10-24
a11y: Keyboard Navigation

dakeyras7:
This PR introduces a consistently designed focus ring and makes the wheelmap completely accessible to keyboard navigation users. You can now reach everything via Tab/Shift-Tab and activate controls via their standard keyboard key.

A couple of things that are worth mentioning:

* In order for mouse input not to activate focus-rings I am using a polyfill for the future `:focus-ring` pseudo selector. I also forked and edited the polyfill to our needs, but a PR for that project will be opened soon.
* I included the map entirely into the tab navigation and made it so, and with @opyh we made it so, that only actually visible markers can be focused. The implementation is not perfect yet and we need to discuss if we should put more time into it. Also the tab order of the markers is not visually sorted  in any way and might jump around.
* The map itself can be panned via the arrow keys once the map has focus. In order to not miss thatI made the focus ring particularly bold.


dakeyras7:
Since #5 contains this branch too, I'm closing this.

PR #25, opened 2018-05-23
Always-visible search bar search bar, pt. I

opyh:
This is base code for the always-visible search bar.

It's only a part of the feature, but new code for the filter toolbar might depend on it, so I made an earlier pull request :)

Still some bugs and clunky look'n'feel (fixing this now), but behavior works generally.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,7 +2,6 @@
 
 import get from 'lodash/get';
 import pick from 'lodash/pick';
-import omit from 'lodash/omit';
 import * as React from 'react';
 import styled from 'styled-components';
 import includes from 'lodash/includes';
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -166,12 +170,26 @@ class FeatureLoader extends React.Component<Props, State> {
...
     updateTouchCapability();
+    this.updateViewportSizeState();
   };
 
+
+  onMoveEndHandler = (state) => {
+    // console.log('Setting query params after moving to', state);
+    // setQueryParams(this.props.history, omit(state, 'bbox'));
+
+    saveState('map.lastZoom', String(state.zoom));
```
mutaphysis: You could combine this into 

```
saveState({
  'map.lastZoom': String(state.zoom),
  'map.lastCenter.lat': String(state.lat),
  …
})
```

PR #27, opened 2018-05-23
Search bar fixes + refactoring

opyh:
This moves the search bar UX closer towards Rainer's new flow designs.

- search bar looks simpler and more button-like now
- search bar now correctly shown at the same time as PoIs on desktops/tablets
- Search input field now sticky at the top
- includes a refactoring to use [`getDerivedStateFromProps` and `componentDidUpdate`](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) instead of legacy React component methods

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -45,6 +45,7 @@ import {
   isWheelmapFeature,
   yesNoLimitedUnknownArray,
   yesNoUnknownArray,
+  getFeatureId,
 } from './lib/Feature';
 
 import { CategoryStrings as EquipmentCategoryStrings } from './lib/EquipmentInfo';
@@ -68,6 +69,7 @@ type Props = {
 
 
 type State = {
+  featureId: ?number | string,
   feature?: ?Feature,
   fetching: boolean,
   toilet: ?string,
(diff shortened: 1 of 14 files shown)


src/App.js
```diff
@@ -96,6 +98,39 @@ type RouteInformation = {
   equipmentInfoId: ?string,
 };
 
+function getRouteInformation(props: Props): ?RouteInformation {
+  const location = props.location;
+  const allowedResourceNames = ['nodes', 'categories', 'search'];
+  const match = location.pathname.match(/(?:\/beta)?\/?(?:(-?\w+)(?:\/([-\w\d]+)(?:\/([-\w\d]+)(?:\/([-\w\d]+))?)?)?)?/i);
```
mutaphysis: Not a fan of this, but I see no better solution immediately, so lets do this :)

src/App.js
```diff
@@ -96,6 +98,39 @@ type RouteInformation = {
...
+
+
+function getFeatureIdFromProps(props: Props): ?string {
+  const { featureId } = getRouteInformation(props) || {};
+  return featureId ? String(featureId) : null;
+}
+
+
+function featureIdHasChanged(newProps: Props, prevState: State) {
+  const result = String(getFeatureIdFromProps(newProps)) !== String(prevState.featureId);
```
mutaphysis: -1: getFeatureIdFromProps already returns a String, please remove the conversions - also `String(null)` is "null"

src/App.js
```diff
@@ -175,13 +213,12 @@ class FeatureLoader extends React.Component<Props, State> {
...
 
   onMoveEndHandler = (state) => {
-    // console.log('Setting query params after moving to', state);
-    // setQueryParams(this.props.history, omit(state, 'bbox'));
-
-    saveState('map.lastZoom', String(state.zoom));
-    saveState('map.lastCenter.lat', String(state.lat));
-    saveState('map.lastCenter.lon', String(state.lon));
-    saveState('map.lastMoveDate', new Date().toString());
+    saveState({
```
mutaphysis: Cool! 👍

src/App.js
```diff
@@ -68,6 +69,7 @@ type Props = {
 
 
 type State = {
+  featureId: ?number | string,
```
mutaphysis: Maybe: Mention null explicitly?

src/App.js
```diff
@@ -198,51 +235,54 @@ class FeatureLoader extends React.Component<Props, State> {
...
-  updateStateFromProps(props: Props) {
-    this.fetchFeature(props);
+  static getDerivedStateFromProps(newProps: Props, prevState: State): State {
+    const result: $Shape<State> = {};
 
-    if (this.featureId(props) !== this.featureId(this.props)) {
-      this.setState({ isFilterToolbarVisible: false });
+    if (featureIdHasChanged(newProps, prevState)) {
+      result.isFilterToolbarVisible = false;
+      result.featureId = getFeatureIdFromProps(newProps);
```
mutaphysis: -1: This will always be a string or null, remove `number` from type in state

src/App.js
```diff
@@ -297,55 +337,24 @@ class FeatureLoader extends React.Component<Props, State> {
...
       if (!feature) return;
-      const currentlyShownId = this.featureId(this.props);
-      const idProperties = [
-        typeof feature.id === 'number' && feature.id,
-        typeof feature._id === 'string' && feature._id,
-        feature.properties && typeof feature.properties.id === 'number' && feature.properties.id,
-        feature.properties && typeof feature.properties._id === 'string' && feature.properties._id
-      ];
-      const fetchedId = String(idProperties.filter(Boolean)[0]);
+      const currentlyShownId = getFeatureId(this.props);
```
mutaphysis: -2: Should this not be `getFeatureIdFromProps`?

opyh: Whoops! Yep!

src/App.js
opyh: Done!

src/App.js
opyh: This grew a bit too long ;)

PR #28, opened 2018-05-24
Ensure that highlights are always visible

mutaphysis:
- Instead of modifying the actual marker, add another
  highlight marker on a separate layer
- This is visible on all zoom levels and not part of
  any clusters
- Cleaned up highlight management a bit


mutaphysis:
screen shot 2018-05-24 at 11 09 44

PR #29, opened 2018-05-28
Show loading indicator for map layer loading events

mutaphysis:
- Extended FetchManager to manage isLoading for layers
- Keeps track of loading events, ties in with
  GlobalActivityIndicator like before
- Some minor flow type fixes

### DIFF
--- a/src/components/MainMenu/GlobalActivityIndicator.js
+++ b/src/components/MainMenu/GlobalActivityIndicator.js
@@ -6,17 +6,17 @@ import { globalFetchManager } from '../../lib/FetchManager';
 
 type State = {
   isShown: boolean;
+  lastError: ?Error;
 }
 
 export default class GlobalActivityIndicator extends React.Component<void, State> {
   state = { isShown: false };
 
   updateState = debounce(() => {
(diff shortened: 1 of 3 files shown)


src/components/MainMenu/GlobalActivityIndicator.js
```diff
@@ -6,17 +6,17 @@ import { globalFetchManager } from '../../lib/FetchManager';
...
   updateState = debounce(() => {
-    this.setState({
-      isShown: globalFetchManager.isLoading(),
-      lastError: globalFetchManager.lastError,
-    });
-  }, 1000);
+    const isShown = globalFetchManager.isLoading();
+    const lastError = globalFetchManager.lastError;
+    this.setState({ isShown, lastError });
+  }, 50, { maxWait: 50, leading: true });
```
opyh: A bit unsure: Shouldn't this be `leading: true, trailing: true`?

opyh: Easy-to-add though if necessary, so no blocker.

PR #34, opened 2018-05-29
 Rewrote map positioning logic

mutaphysis:
- feature > lat & lon
- use padding
- move map when selection is not inside padding
- fixed problems with position state management


mutaphysis:
Hi @opyh - I think there is more potential, my feeling is that the Map is still mixing both a stateful component and at the same time it is being controlled by the App. All the map center/zoom logic should reside in App js. I solved it like that in wheelmap-pro, and was a bit saner. 

But this works kind of nice.

opyh:
You are right, Map.js does too many things at once in the meantime. I hope there's a chance to reuse the new logic from WMP at some point.

PR #36, opened 2018-05-29
Close modals on click outside modal

opyh:
This adds a fullscreen backdrop modal component with subtle 3D effect, to be used when you want to display something modally.

To be compatible to all platforms, it's not using `-webkit-backdrop-filter`, but encapsulates everything to be blurry behind the modal dialog in a `<div>` with a `filter` CSS property.

It's not perfectly generic, but works for all our states so far, unifying their look :)

The new code also moves some component state (isReportMode/isEditMode) to the outside of the node toolbars to simplify things.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -675,6 +675,7 @@ class FeatureLoader extends React.Component<Props, State> {
     const nodeToolbar = this.renderNodeToolbar({ isNodeRoute, featureId, equipmentInfoId, isEditMode, isReportMode });
 
     return (<div className={classList.join(' ')}>
+      {!isMainMenuInBackground && mainMenu}
       <div className="behind-backdrop">
         {isMainMenuInBackground && mainMenu}
         {isLocalizationLoaded && this.renderSearchToolbar({ isInert: searchToolbarIsInert, category, searchQuery, lat, lon })}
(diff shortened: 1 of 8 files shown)


opyh:
Whoops, this breaks the main menu on desktop. Please don't merge yet…

opyh:
Fixed, could you have a look?

src/App.js
```diff
@@ -559,23 +569,61 @@ class FeatureLoader extends React.Component<Props, State> {
...
+      (this.state.isFilterToolbarVisible && this.state.isLocalizationLoaded) ||
+      this.state.isMainMenuOpen ||
+      this.state.isOnboardingVisible ||
+      this.state.isNotFoundVisible ||
+      isEditMode ||
+      this.state.isReportMode;
+
+    return <FullscreenBackdrop
+      onClick={() => {
+        console.log('backdrop clicked');
```
mutaphysis: NitPick: Unneeded

src/App.js
```diff
@@ -559,23 +569,61 @@ class FeatureLoader extends React.Component<Props, State> {
...
+  renderFullscreenBackdrop() {
+    const routeInformation = getRouteInformation(this.props);
+    const { isEditMode } = routeInformation || {};
+    const isActive =
+      (this.state.isFilterToolbarVisible && this.state.isLocalizationLoaded) ||
+      this.state.isMainMenuOpen ||
+      this.state.isOnboardingVisible ||
+      this.state.isNotFoundVisible ||
+      isEditMode ||
+      this.state.isReportMode;
```
mutaphysis: 0: Move to function?

opyh: Yep! Will do that in the next PR, where I had to reorder some of these things.

PR #37, opened 2018-05-29
Add "show on osm" to all POIs

mutaphysis:
Maybe all is a bit much? Having this on HERE POIs is a bit weird.

### DIFF
--- a/src/components/NodeToolbar/NodeToolbar.js
+++ b/src/components/NodeToolbar/NodeToolbar.js
@@ -11,6 +11,7 @@ import styled from 'styled-components';
 import { Dots } from 'react-activity';
 import Categories from '../../lib/Categories';
 import type { Category } from '../../lib/Categories';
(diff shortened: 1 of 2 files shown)


opyh:
An argument for it: each mapping app out there has its advantages and disadvantages for specific use cases. The more possibilities in this case, the better – maybe we also learn something about which use case works better in which mapping app. Would try it live before restricting it intentionally :)

src/components/NodeToolbar/NodeToolbar.js
```diff
@@ -271,6 +272,11 @@ class NodeToolbar extends React.Component<Props, State> {
     const reportButtonCaption = t`Report an Issue`;
     const placeName = placeNameFor(get(this.props, 'feature.properties'), this.state.category);
 
+
+    // translator: Button caption shown in the place toolbar
+    const showOnOsmCaption = t`Show on OpenStreetMap`;
```
opyh: Could you use 'Open' instead of 'Show'?

PR #45, opened 2018-06-12
New filter UX

opyh:
Sorry that this is such a big chunk – this was affecting a lot of places in the code.

This implements the new filter UX from Rainer (with one difference: There is no 'all places' category, which made many things easier). This feels alright for me, and much better than before :)


package.json
```diff
@@ -115,13 +116,13 @@
...
+    "start": "REACT_APP_TAGS_TO_APPEND_TO_INDEX_HTML_HEAD=\"\" extract=1 node scripts/start.js",
     "start-cordova": "REACT_APP_ENV=cordova REACT_APP_WHEELMAP_API_BASE_URL=https://wheelmap.org PUBLIC_URL=\"./\"",
     "analyze": "source-map-explorer build/static/js/main.*",
     "test": "node scripts/test.js --env=jsdom",
-    "build": "extract=1 BABEL_DISABLE_CACHE=1 node scripts/build.js",
-    "build-cordova": "REACT_APP_ENV=development REACT_APP_WHEELMAP_API_BASE_URL=https://wheelmap.org PUBLIC_URL=\"./\" node scripts/build.js; rm -rf www; cp -r build www",
-    "build-staging": "REACT_APP_ENV=staging node scripts/build.js",
+    "build": "extract=1 BABEL_DISABLE_CACHE=1 REACT_APP_TAGS_TO_APPEND_TO_INDEX_HTML_HEAD=\"\" node scripts/build.js",
+    "build-cordova": "REACT_APP_TAGS_TO_APPEND_TO_INDEX_HTML_HEAD=\"<script src='cordova.js'></script>\" REACT_APP_ENV=development REACT_APP_WHEELMAP_API_BASE_URL=https://wheelmap.org PUBLIC_URL=\"./\" node scripts/build.js; rm -rf www; cp -r build www",
+    "build-staging": "REACT_APP_TAGS_TO_APPEND_TO_INDEX_HTML_HEAD=\"\" REACT_APP_ENV=staging node scripts/build.js",
```
opyh: This stuff is to insert the cordova script tag only for cordova builds.

src/App.js
```diff
@@ -331,433 +266,258 @@ class FeatureLoader extends React.Component<Props, State> {
...
-  }
-
-
-  fetchFeature(featureId: ?string): void {
-    if (!featureId) {
-      this.setState({ featureId: null, feature: null, lat: null, lon: null, zoom: null });
-      return;
-    }
-    this.setState({ fetching: true, featureId });
+  fetchFeature(featureId: string): void {
```
opyh: This does not need to check for a `fetching` flag anymore – it's should only be called when actually needed.

src/components/SearchToolbar/AccessibilityFilterMenu.js
```diff
@@ -0,0 +1,216 @@
...
+            <AccessibilityFilterButton
+              accessibilityFilter={availableFilters[key].accessibilityFilter}
+              toiletFilter={availableFilters[key].toiletFilter}
+              caption={availableFilters[key].caption}
+              category={category}
+              isMainCategory
+              isActive={currentFilterKey}
+              showCloseButton={shownFilterKeys.length === 1}
+              onKeyDown={({nativeEvent}) => {
+                const tabPressedOnLastButton = index === lastIndex && nativeEvent.key === 'Tab' && !nativeEvent.shiftKey;
```
mutaphysis: Maybe: move into its own method?

src/components/SearchToolbar/AccessibilityFilterModel.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import isEqual from 'lodash/isEqual';
```
mutaphysis: not needed?

src/components/SearchToolbar/CategoryButton.js
```diff
@@ -1,64 +1,163 @@
...
+        font-size: 0.80em;
+        margin-top: 0.5em;
+        color: ${colors.darkSelectedColor};
       }
     }
   }
 `;
 
 
+function urlForFilters({ history, id, accessibilityFilter, toiletFilter, showCloseButton }) {
```
mutaphysis: This method exists twice with slightly different implementations, can they be consolidated?

src/components/SearchToolbar/SearchResult.js
```diff
@@ -0,0 +1,265 @@
+// @flow
```
mutaphysis: The file name/class is a bit confusing, it contains the whole routing logic - more than just search - , but I wouldn't have guessed.

opyh: Agree, best would be to inject a function from the outside that creates the necessary `href` for a feature, from <App /> all the way down – what do you think?

src/components/SearchToolbar/AccessibilityFilterMenu.js
opyh: Even better: Not needed anymore because tab handling does the right thing here.

src/components/SearchToolbar/SearchResult.js
opyh: Made a ticket for this: https://trello.com/c/49XYKKDi/758-side-quest-inject-href-function-for-searchresult-from-outside

PR #51, opened 2018-06-26
Improve main view focus behavior

opyh:
This re-introduces the `manageFocus` method. Its implementation is far from complete but this fixes a few small cases.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -394,6 +394,7 @@ class Loader extends React.Component<Props, State> {
   onMapClick = () => {
     this.closeSearch();
     this.setState({ isMainMenuOpen: false, isSearchToolbarExpanded: false });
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -418,7 +419,7 @@ class Loader extends React.Component<Props, State> {
       isReportMode: false,
     });
     if (this.isEditMode()) {
-      this.props.history.push(`/nodes/${this.state.featureId}`);
+      this.props.history.push(`/nodes/${String(this.state.featureId)}`);
```
mutaphysis: Is this required? I though template literals would do string conversion when doing string interpolation?

opyh: Flow complains without this – strictly, `featureId` could be `undefined` which formally is not castable to String.

PR #54, opened 2018-06-26
Blur search bar focus when it is closed, fix more focus management

opyh:
This ensures the focus is correctly set when node/search toolbars (dis-)appear.

### DIFF
--- a/src/MainView.js
+++ b/src/MainView.js
@@ -75,7 +75,7 @@ type Props = {
   onResetCategory: (() => void),
   onClickSearchToolbar: (() => void),
   onCloseSearchToolbar: (() => void),
-  onClickSearchButton: ((event: UIEvent) => void),
+  onClickSearchButton: (() => void),
   onCloseNotFoundDialog: (() => void),
   onToggleMainMenu: (() => void),
   onClickFullscreenBackdrop: (() => void),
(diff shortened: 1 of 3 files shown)


src/MainView.js
```diff
@@ -276,7 +276,7 @@ class MainView extends React.Component<Props, State> {
 
   renderSearchButton() {
     return <SearchButton
-      onClick={this.props.onClickSearchButton}
+      onClick={(event) => setTimeout(() => this.props.onClickSearchButton(event), 10)}
```
mutaphysis: a bit smelly. Maybe add a comment?

src/MainView.js
```diff
@@ -75,7 +75,7 @@ type Props = {
   onResetCategory: (() => void),
   onClickSearchToolbar: (() => void),
   onCloseSearchToolbar: (() => void),
-  onClickSearchButton: (() => void),
+  onClickSearchButton: ((event: UIEvent) => void),
```
mutaphysis: Is the event even still required? I cannot find anything in this change. Its weird that one method is different, and also weird that the internal implementation details bleed out of the component

opyh: Yeah, <App /> uses it for `stopPropagation`. Probably it's better to stop the propagation inside the component, you are right.

PR #56, opened 2018-07-02
Expand node toolbar when opening a PoI on mobile

opyh:
This expands the toolbar to cover around half of the screen when opening the first PoI.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -312,22 +312,17 @@ class Loader extends React.Component<Props, State> {
     const searchToolbarDidAppear = isSearchToolbarDisplayed && !wasSearchToolbarDisplayed;
 
     if ((nodeToolbarDidDisappear || searchToolbarDidDisappear)) {
-      console.log('Focusing', this.lastFocusedElement);
       window.document.activeElement.blur();
       if (this.lastFocusedElement) this.lastFocusedElement.focus();
     }
 
     if ((nodeToolbarDidAppear || featureIdHasChanged) && this.mainView.nodeToolbar) {
       this.lastFocusedElement = document.activeElement;
-      console.log('Saving last focused element:', this.lastFocusedElement);
-      console.log('Focusing', this.mainView.nodeToolbar);
       this.mainView.nodeToolbar.focus();
     }
 
     if (searchToolbarDidAppear && this.mainView.searchToolbar) {
       this.lastFocusedElement = document.activeElement;
-      console.log('Saving last focused element:', this.lastFocusedElement);
-      console.log('Focusing', this.mainView.searchToolbar);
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -312,22 +312,17 @@ class Loader extends React.Component<Props, State> {
...
       this.lastFocusedElement = document.activeElement;
-      console.log('Saving last focused element:', this.lastFocusedElement);
-      console.log('Focusing', this.mainView.nodeToolbar);
       this.mainView.nodeToolbar.focus();
     }
 
     if (searchToolbarDidAppear && this.mainView.searchToolbar) {
       this.lastFocusedElement = document.activeElement;
-      console.log('Saving last focused element:', this.lastFocusedElement);
-      console.log('Focusing', this.mainView.searchToolbar);
```
mutaphysis: :)

src/components/Toolbar.js
```diff
@@ -174,11 +189,14 @@ class Toolbar extends React.Component<Props, State> {
         result = stop;
       }
     });
+    console.log('Nearest offset for preferred offset', topOffset, 'is', result, '- stops:', stops);
```
mutaphysis: Is this triggered often?

src/components/Toolbar.js
```diff
@@ -190,11 +208,13 @@ class Toolbar extends React.Component<Props, State> {
...
+  /** @returns An array of top position offsets that the toolbar is allowed to stop on. */
+
   getStops(): number[] {
-    const minimalTopPosition = this.getMinimalTopPosition();
+    const maximalTopPosition = this.getMaximalTopPosition();
     const stops = uniq([
-      minimalTopPosition,
-      Math.max(minimalTopPosition, Math.floor(this.state.viewportSize.height / 2)),
+      maximalTopPosition,
+      Math.max(maximalTopPosition, Math.floor(this.state.viewportSize.height / 2)),
```
mutaphysis: Should this not be Math.min? Now it returns always something larger or equal to maximalTopPosition

src/components/Toolbar.js
```diff
@@ -211,7 +231,7 @@ class Toolbar extends React.Component<Props, State> {
     const lastTopOffset = this.state.lastTopOffset;
 
     let topOffset = this.state.topOffset || this.state.lastTopOffset;
-    topOffset = Math.max(this.getMinimalTopPosition(), topOffset);
+    topOffset = Math.max(this.getMaximalTopPosition(), topOffset);
```
mutaphysis: Same as above

opyh:
Logic is correct – naming was misleading though ;) Fixed.

PR #62, opened 2018-07-04
Improve location UX

opyh:
- When opening the app, start locating by default
- Tapping/Clicking once the locate-me button when location is known pans the map to device location
- Tapping the locate-me button and then tapping on the map does not reset the map position to previous position anymore
- After a while after starting the app, the API delivers a new location, leading to a map pan
- When you start the app with empty cache and without known location, it shows the Cologne dom
- When you start the app on a specific PoI with a saved location, the saved location is used until the PoI location is loaded
- When a PoI is shown, the map does not automatically pan somewhere else anymore when a new device location is known
- Show help when you disallow location services in the privacy settings


src/components/Map/Map.js
```diff
@@ -281,6 +282,39 @@ export default class Map extends React.Component<Props, State> {
...
+    }
+  }
+
+  setupLocateMeButton(map: L.Map) {
+    this.locateControl = addLocateControlToMap(map, {
+      locateOnStart: this.props.locateOnStart || false,
+      onLocationError: (error: any) => {
+        if (error && error.type && error.type === 'locationerror' && error.code && error.code === 1) {
+          // System does not allow to use location services
+          debugger
```
mutaphysis: Please remove

src/components/Map/NotificationButton.js
```diff
@@ -0,0 +1,135 @@
...
+    color: white;
+    z-index: 1;
+    border-radius: 4px;
+    font-size: .9rem;
+    font-weight: 300;
+    text-shadow: 0 0px 1px rgba(0,0,0,0.5);
+  }
+
+
+  /* only show nose with zoom buttons */
```
mutaphysis: comment might be slightly out of date

PR #71, opened 2018-07-17
Cleanup & fix origin policies

mutaphysis:
Fixes several allow origin issues and cleans up Content-Security-Policy.

- accessibility cloud was accessed without the www.
- make URLDataCache use cordova fetch
- use env variables in index.html instead of hardcoded urls


opyh:
Thank you very much for this!

PR #77, opened 2018-07-19
Feat/restyle feature details 2

opyh:


src/components/NodeToolbar/AccessibilityEditor/WheelchairStatusEditor.js
```diff
@@ -217,7 +209,7 @@ class WheelchairStatusEditor extends React.Component<Props, State> {
           name="accessibility"
           selectedValue={wheelchairAccessibility}
           onChange={(newValue) => { this.setState({ wheelchairAccessibility: newValue }); }}
-          className={`${wheelchairAccessibility} ${valueIsDefined ? 'has-selection' : ''} radio-group`}
+          className={`${String(wheelchairAccessibility)} ${valueIsDefined ? 'has-selection' : ''} radio-group`}
```
mutaphysis: -0.5: `${String(wheelchairAccessibility)}` -> `String(null) = 'null'`  `String(undefined) = 'undefined')` 

Maybe `${wheelchairAccessibility || ''}` if your goal was to prevent class="undefined …"

src/components/NodeToolbar/AccessibilitySection/StyledFrame.js
```diff
@@ -0,0 +1,63 @@
...
+
+  &:before {
+    display: block;
+    position: absolute;
+    content: " ";
+    top: -8px;
+    left: 1em;
+    width: 12px;
+    height: 8px;
+    background: url(../NoseDetailsContainer.svg) no-repeat; /* TODO: decide on outline vs. color & add nose */
```
mutaphysis: 0: This does not work under cordova, atm. But it was already like this before

src/components/NodeToolbar/AccessibilityEditor/WheelchairStatusEditor.js
opyh: I was thinking of this case and thought having 'null' or 'undefined' as class there would make debugging easier if it happens… Or is this a bad idea? :)

src/components/NodeToolbar/AccessibilitySection/StyledFrame.js
opyh: Yep! I've created a ticket for this: https://trello.com/c/9gIa1XlT/903-add-little-arrow-to-place-a11y-box

src/components/NodeToolbar/AccessibilityEditor/WheelchairStatusEditor.js
mutaphysis: Yes, it's a bad idea. First it might of course select an unwanted class named 'undefined'. Second undefined is a valid value (at least according to the flow type, so it should be treated correctly. But maybe 'undefined' is the correct treatment then, if a selector for such a class would exist.

PR #79, opened 2018-07-19
Fix iOS release deployment

opyh:
This includes necessary fixes for iOS beta/release builds. I'll rebase this when #77 is merged.


fastlane/Matchfile
```diff
@@ -1,6 +1,5 @@
 git_url("git@github.com:sozialhelden/certificates.git")
 
-type("development") # The default type, can be: appstore, adhoc, enterprise or development
```
opyh: This is set in `Fastfile` as parameter to `match()` instead

PR #80, opened 2018-07-19
Fix toolbar scrolling and styling

opyh:
This fixes styling issues with toolbars and improves iOS scrolling behavior. It's not perfect as it removes native touch scroll behavior for now, but works much better than before. I did not test this on Android yet.


src/components/Toolbar.js
```diff
@@ -238,14 +263,30 @@ class Toolbar extends React.Component<Props, State> {
...
   }
 
 
+  cancelTouchIfMoving = (event: TouchEvent) => {
+    const isScrollable = this.isAtTopmostPosition();
+    if (this.props.onScrollable) {
+      this.props.onScrollable(isScrollable);
+    }
+    if (isScrollable) {
+      console.log('Did not prevent scrolling');
```
mutaphysis: remove

src/components/Toolbar.js
```diff
@@ -238,14 +263,30 @@ class Toolbar extends React.Component<Props, State> {
...
+      return;
+    }
+    event.preventDefault();
+    console.log('Prevented scrolling');
```
mutaphysis: remove

PR #87, opened 2018-07-20
Unify a11y editors

opyh:
This unifies the code + layout for both wheelchair+toilet accessibility editor dialogs.

Also done here:

- Back button in a11y editors now using `preventDefault` to fix double triggering
- Split strings for imperial units and metric units. Translators will have to translate 2 strings, but users will get a string in their own measuring system only which should make things simpler for them.
- Place icon now stays where it is in toilet and report dialogs. The wheelchair a11y editor does not show it to not confuse people who use it
- More removed lines than added lines \o/


src/components/NodeToolbar/AccessibilityEditor/RadioStatusEditor.js
```diff
@@ -0,0 +1,271 @@
...
+      (properties.node_type && properties.node_type.identifier) || properties.category;
+
+    if (!categoryId) {
+      return;
+    }
+
+    Categories.getCategory(categoryId).then(() => {
+      const iconId = getIconNameForProperties(properties);
+      this.setState({categoryId: iconId});
+    })
```
mutaphysis: NIT: Missing ;

src/components/NodeToolbar/AccessibilityEditor/RadioStatusEditor.js
```diff
@@ -0,0 +1,271 @@
...
+
+
+  onRadioGroupKeyDown = ({nativeEvent}) => {
+    if(nativeEvent.key === 'Enter') {
+      this.onSaveButtonClick();
+    }
+  }
+
+
+  onSaveButtonClick = () => {
```
mutaphysis: Prevent propagation here as well?

src/components/NodeToolbar/AccessibilityEditor/RadioStatusEditor.js
```diff
@@ -0,0 +1,271 @@
...
+    const valueIsDefined = selectedValue !== 'unknown';
+
+    return <StyledRadioGroup
+      name="accessibility"
+      selectedValue={selectedValue}
+      onChange={(newValue) => { this.setState({ selectedValue: newValue }); }}
+      className={`${String(selectedValue)} ${valueIsDefined ? 'has-selection' : ''} radio-group`}
+      onKeyDown={this.onRadioGroupKeyDown}
+      role="radiogroup"
+      aria-label={t`Wheelchair Accessibility`}
```
mutaphysis: Add translator info?

src/components/NodeToolbar/AccessibilityEditor/RadioStatusEditor.js
```diff
@@ -0,0 +1,271 @@
...
+
+  renderRadioGroup() {
+    const { selectedValue } = this.state;
+    const valueIsDefined = selectedValue !== 'unknown';
+
+    return <StyledRadioGroup
+      name="accessibility"
+      selectedValue={selectedValue}
+      onChange={(newValue) => { this.setState({ selectedValue: newValue }); }}
+      className={`${String(selectedValue)} ${valueIsDefined ? 'has-selection' : ''} radio-group`}
```
mutaphysis: -1: Still same as last time, I guess 👯‍♂️

src/components/NodeToolbar/AccessibilityEditor/RadioStatusEditor.js
```diff
@@ -0,0 +1,271 @@
...
+      // prevent clicking the next close button as well
+      event.preventDefault();
+      event.stopPropagation();
+    }
+  }
+
+
+  renderCloseButton() {
+    return !this.props.inline && <CloseLink
+      className='close-link'
```
mutaphysis: -1: Unneeded, already in CloseLink component

src/components/NodeToolbar/AccessibilityEditor/WheelchairStatusEditor.js
```diff
@@ -2,285 +2,39 @@
...
+    {...props}
+    undefinedStringValue="unknown"
+    getValueFromFeature={feature => feature.properties.wheelchair}
+    saveValue={(value) => saveWheelchairStatus({ ...props, value })}
+    renderChildrenForValue={({ value, categoryId }) => <Icon accessibility={value} category={categoryId} size='medium' withArrow shadowed centered />}
+    shownStatusOptions={['yes', 'limited', 'no']}
+    captionForValue={value => shortAccessibilityName(value)}
+    descriptionForValue={value => accessibilityDescription(value)}
+  >
+    <header id="wheelchair-accessibility-header">{t`How wheelchair accessible is this place?`}</header>
```
mutaphysis: Add translator note

src/components/NodeToolbar/AccessibilityEditor/saveStatus.js
```diff
@@ -0,0 +1,63 @@
...
+
+  return fetch(url, requestOptions)
+    .then(() => {
+      setTimeout(() => {
+        if (typeof options.onClose === 'function') options.onClose();
+        if (typeof options.onSave === 'function') options.onSave(value);
+      }, 50);
+    })
+    .catch((e) => {
+      window.alert(t`Sorry, could not mark this place because of an error: ${e}`);
```
mutaphysis: Just noticed now, alerts are not really beautiful.
Add translators note?

src/components/NodeToolbar/AccessibilityEditor/saveStatus.js
```diff
@@ -0,0 +1,63 @@
...
+    });
+}
+
+
+export function saveToiletStatus(options: ExternalSaveOptions<YesNoUnknown>) {
+  const url = `${config.wheelmapApiBaseUrl}/nodes/${options.featureId}/update_toilet.js`;
+  return save({ ...options, url, propertyName: 'wheelchair_toilet', jsonPropertyName: 'toilet' });
```
mutaphysis: So this can also save toilet=no now?

src/components/NodeToolbar/NodeToolbar.js
```diff
@@ -185,7 +185,7 @@ class NodeToolbar extends React.Component<Props, State> {
       category={category}
       parentCategory={parentCategory}
       onClickCurrentMarkerIcon={onClickCurrentMarkerIcon}
-      showOnlyBasics={!!this.props.modalNodeState}
+      hasIcon={!this.props.modalNodeState || includes(['edit-toilet-accessibility', 'report'], this.props.modalNodeState)}
```
mutaphysis: -1: extract long boolean into variable

src/lib/useImperialUnits.js
```diff
@@ -0,0 +1,5 @@
+import { currentLocales } from './i18n';
+
+export default function useImperialUnits() {
+  return currentLocales[0] === 'en' || Boolean(currentLocales[0].match(/(UK|US)$/));
```
mutaphysis: This could also be used on the leaflet map.
-1: Canada seems to be mostly using inch for length measurements - see https://en.wikipedia.org/wiki/Metrication_in_Canada#Engineering and following

opyh: 'Canadians are exposed to both metric and imperial units, and it is not unusual for there to be references to both metres and feet, acres and hectares, or grams and ounces, in the same conversation.' suggests we need an additional flag for countries that use both 😅Added CA as imperial now.

src/components/NodeToolbar/NodeToolbar.js
opyh: Good idea!

src/components/NodeToolbar/AccessibilityEditor/saveStatus.js
opyh: Yes, it can – Did I miss something? Not sure what you mean with this…

PR #93, opened 2018-08-08
Added retranslate scripts

mutaphysis:
Take a look at an example replacement in https://github.com/sozialhelden/wheelmap-react-frontend/pull/94/commits/a1274dbfdb40a89ab3d82fc56990fc04545432a7


retranslate/.babelrc
```diff
@@ -0,0 +1,12 @@
+{
+  "plugins": [
+    "syntax-class-properties",
+    "syntax-object-rest-spread",
+    "syntax-jsx",
+    "syntax-flow",
+    ["./retranslate/babel-plugin-retranslate/dist-plugin.js", {
+      "replacement": "../public/i18n/de.txt"
```
opyh: I guess this should be `en_US.txt`?

retranslate/rewrite-pos/rewrite-pos.js
```diff
@@ -0,0 +1,74 @@
...
+      } else {
+        console.warn(`new source translation missing for '${newValue.msgid}'`);
+        rewrittenContext[key] = newValue;
+      }
+
+      // retranslate plurals
+      if (newValue.msgid_plural) {
+        if (rewriteValue.msgstr[1]) {
+          const newKey = rewriteValue.msgstr[1];
+          newValue.msgid_plural = newKey;
```
opyh: Does this work with languages that have more than two plural forms, e.g. Russian?

retranslate/run.sh
```diff
@@ -0,0 +1,37 @@
+# setup environment
+export NEW_SOURCE_PO=${NEW_SOURCE_PO:-'./public/i18n/en_US.txt'}
+export NODE_ENV=development
+
+# ensure i18n & src folder are reset 
+git checkout -- public/i18n/ src
```
opyh: I'd be afraid that this script accidentally my changes.

image

How about if this exits if there are changes and lets the user do the cleanup themselves?

Maybe like this?

```bash
test -z "$(git status --porcelain)" || (echo "Your repo must be clean"; exit 1)
```

retranslate/rewrite-pos/rewrite-pos.js
mutaphysis: Only the first plural form is used as the new plural id, this should not matter for the other plural forms

retranslate/.babelrc
mutaphysis: Will fix, and add script to package json

retranslate/run.sh
mutaphysis: Oh yes, this was more for dev, will add

PR #95, opened 2018-08-08
Retranslated to en_US

mutaphysis:


opyh:
This is included in https://github.com/sozialhelden/wheelmap-react-frontend/pull/96 – thanks anyway for creating it! :)

PR #101, opened 2018-08-30
Release/ios/2018 08 29

opyh:
This contains last bugfixes before submission to App Store / Play Store. Will rebase this when the other PRs are merged so it's easier to review.

Fixes:
- Wrong layout on iPads fixed, now there is a padding between the combi search button and place details (it was missing before)
- On iOS/Android, the detailed accessibility dialog that is shown after tapping the big a11y buttons of a gray place now is preset correctly (the URL parameter was not interpreted because MemoryRouter is used in Cordova)
- Changing a place's accessibility now updates the accessibility in the feature details always (before it worked only in 50% of the cases)
- Toilet dialog not opened automatically anymore after editing general wheelchair accessibility
- Added missing `/beta` prefixes to several URLs where it was missing
- Switched off scheme autocreation to ensure cordova always creates a build. Without this, if there is no matching scheme, cordova Xcode builds just exit cleanly even if no build has been created, leading to old binaries being used in tasks that happen afterwards.


mutaphysis:
-1: There is one `this.props.history.push(`/nodes/${featureId}`);` left in the code (`src/components/NodeToolbar/NodeToolbar.js:187` - why do we actually need this at all?

opyh:
@mutaphysis It's not necessary when using <Link>. I have updated it.

PR #104, opened 2018-09-11
Set view only once on start when new location is found

opyh:
I've tested this on iOS simulator and on Android device – after the fix, the location is not jumping anymore when a new location event arrives.

The new code does `setView` itself on start if necessary, but only once. Before, it was setting `options.setView = true` on the Leaflet locate control, which meant that every new location sent by the device was causing a jump if the location button was active when restarting the app.

This code does not touch the behavior of panning the map after clicking the locate-me button once.

### DIFF
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -95,8 +95,11 @@ export default class Map extends React.Component<Props, State> {
   accessibilityCloudTileLayer: ?GeoJSONTileLayer;
   highLightLayer: ?L.Layer;
   locateControl: ?LeafletLocateControl;
+  mapHasBeenMoved: boolean = false;
 
   onMoveEnd() {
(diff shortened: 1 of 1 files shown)


src/components/Map/Map.js
```diff
@@ -277,6 +277,12 @@ export default class Map extends React.Component<Props, State> {
         }
       }
     });
+
+    map.once('locationfound', () => {
+      if (this.locateControl && this.props.locateOnStart) {
```
mutaphysis: Would this still override the position of the map if the user dragged already?

lennerd:
What is the desired behaviour when the users clicks on the locate-me button and the location changes without any drags, searches, etc.? Should the map center move to the new location marker positions until some interaction is happening (search, drag)?

PR #105, opened 2018-09-11
WIP testing documentation

mutaphysis:
- Got bored, but maybe its already a bit helpful


dakeyras7:
This almost reads to me like an e2e test. Wouldn't it be nice to translate this into one somehow? In my last project I really liked using https://www.cypress.io.

mutaphysis:
Yes, I also thought that automating this would be super nice :)

PR #106, opened 2018-09-11
Reformat all javascript files

mutaphysis:
Based on #102


mutaphysis:
Why do we need a package-lock.json, when we are using yarn?

lennerd:
@opyh Pointed out that we cannot use `npm audit` without a `package-lock.json` file.

mutaphysis:
@lennerd thanks, I get it!

PR #110, opened 2018-09-12
Fix sorting bug of accessiblity and toilet status filters.

lennerd:
The Map component was sorting status and toilet status filters when comparing arrays for equality. This sorting was mutable and changing the order of given status filters for other components too. The StatusFilter component used the equality of status filters to see if a status filter was selected by user. The changed order broke this comparison.

This PR moves the sorting of accessibility and toilet status filters to the App component (getDerivedStateFromProps) so that all components always have the sorted list of statuses.

See: https://trello.com/c/DOl6ELoV

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,8 @@
     "ua-parser-js": "^0.7.18",
     "underscore.string": "^3.3.4",
     "unfetch": "^3.0.0",
-    "wicg-focus-ring": "https://github.com/dakeyras7/focus-ring#include-enter-keydown"
+    "wicg-focus-ring": "https://github.com/dakeyras7/focus-ring#include-enter-keydown",
+    "cordova-android": "~7.0.0"
   },
   "devDependencies": {
     "@babel/polyfill": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -55,8 +55,7 @@
     "ua-parser-js": "^0.7.18",
     "underscore.string": "^3.3.4",
     "unfetch": "^3.0.0",
-    "wicg-focus-ring": "https://github.com/dakeyras7/focus-ring#include-enter-keydown",
-    "cordova-android": "~7.0.0"
+    "wicg-focus-ring": "https://github.com/dakeyras7/focus-ring#include-enter-keydown"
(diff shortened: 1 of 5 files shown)


lennerd:
@opyh Maps now also sort incoming currently active status filters like discussed in Slack.

package.json
```diff
@@ -55,7 +55,8 @@
     "ua-parser-js": "^0.7.18",
     "underscore.string": "^3.3.4",
     "unfetch": "^3.0.0",
-    "wicg-focus-ring": "https://github.com/dakeyras7/focus-ring#include-enter-keydown"
+    "wicg-focus-ring": "https://github.com/dakeyras7/focus-ring#include-enter-keydown",
+    "cordova-android": "~7.0.0"
```
mutaphysis: Is this a non-dev dependency or can we move it to all the other cordova things?

mutaphysis: yarn.lock was not updated, though, was this intentional?

src/App.js
```diff
@@ -93,15 +93,29 @@ type State = {
 
 function getAccessibilityFilterFrom(statusString: string): YesNoLimitedUnknown[] {
   const allowedStatuses = yesNoLimitedUnknownArray;
-  if (!statusString) return [].concat(allowedStatuses);
-  const result = statusString.split(/\./).filter(s => includes(allowedStatuses, s));
+
+  // Safe mutable sort as filter always returns a new array.
+  const result = statusString
```
mutaphysis: MINOR: This could be moved into a small helper function, and used also below.

package.json
lennerd: You are right. This was added automatically when building for android via `yarn build-android` I guess.

mutaphysis: @opyh mentioned that cordova tends to reorder the package.json :/

lennerd: 🙄 I will remove it.

src/App.js
lennerd: Good catch!

lennerd:
Added more DRYness and removed the cordova dependency.

PR #111, opened 2018-09-12
Improve report button a11y + lightbox

opyh:
- align the font sizes of pagination and report button in the photo lightbox
- remove duplicate close button in the lightbox
- Improve button + pagination readability by using a darker background color
- fix a few translator comments

This fixes https://trello.com/c/8IukzJd8/979-report-button-on-gallery-should-be-clickable-visible.

Note that report buttons are not visible for all places. If you want to see it, here is a place that has a reportable image: http://localhost:3000/beta/nodes/554200132

### DIFF
--- a/src/components/NodeToolbar/Photos/PhotoSection.js
+++ b/src/components/NodeToolbar/Photos/PhotoSection.js
@@ -175,7 +175,6 @@ class PhotoSection extends React.Component<Props, State> {
           onClick={this.reportImage}
           className="report-image"
         >{t`Report image`}</button>
-        <button onClick={this.closeLightbox} className="close-lightbox">{t`Close`}</button>
       </section>,
     ];
   };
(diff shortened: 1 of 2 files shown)


src/components/NodeToolbar/Photos/PhotoSection.js
```diff
@@ -201,14 +200,15 @@ class PhotoSection extends React.Component<Props, State> {
           isOpen={this.state.isLightboxOpen}
           // translator: divider between <currentImageIndex> and <imageCount> in lightbox, such as 1 of 10
           imageCountSeparator={' ' + t`of` + ' '}
-          // tranlator: alt info on next image button in lightbox
+          // translator: alt info on next image button in lightbox
```
mutaphysis: Good catch!

PR #112, opened 2018-09-12
Bugfixes: Improve main menu

opyh:
- Fix main menu button in landscape iPad
- allow Escape button in main menu on any button

This fixes https://trello.com/c/3HSy05gY/972-opening-main-menu-on-ipad-landscape-opens-wheelmap-news.

### DIFF
--- a/src/components/MainMenu/MainMenu.js
+++ b/src/components/MainMenu/MainMenu.js
@@ -92,9 +92,10 @@ class MainMenu extends React.Component<Props, State> {
     }
   }
 
-  toggleMenu() {
+  toggleMenu = (event: Event) => {
     this.props.onToggle(!this.props.isOpen);
-  }
+    event.preventDefault();
+  };
 
   setupFocusTrap() {
     if (this.homeButton && this.addPlaceLink) {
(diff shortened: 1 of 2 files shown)


src/components/MainMenu/MainMenu.js
```diff
@@ -193,6 +200,7 @@ class MainMenu extends React.Component<Props, State> {
             ref={homeButton => (this.homeButton = homeButton)}
             tabIndex={hideFromFocus ? -1 : 0}
             aria-label={t`Home`}
+            onKeyDown={this.handleKeyDown}
```
mutaphysis: Keydown bubbles, shouldn't one handler on the container be enough? 

https://developer.mozilla.org/en-US/docs/Web/Events/keydown

opyh: This is what I did first (also because it's more DRY), but the linter said that `<div>`s should not have a keydown handler. You're right though, I should ignore the linter rules here and add a `eslint-ignore` statement as the linter rule is meant to help for a different case. Will change it!

dakeyras7: My a11y-50-cent here: The rule probably applied because non-interactive elements like `<div>`s should not be used instead of a proper interactive HTML tag. But in cases we have to for some reason, it's probably best to assign an aria-role to that `<div>` then conveys the actual role of it. Not sure, what is needed here.

mutaphysis: Let's take it as build by @opyh

PR #113, opened 2018-09-13
Fix routing for 404 not found.

lennerd:
The `routeInformation` is `null` when the given route does not match the URL format. Detecting a 404 based on the URL format might not be a save way of detecting a 404 but should be sufficient for a first version.

Also I think the 404 error page looks a bit broken on desktop. Not sure if I should fix this with this PR, too?

screenshot 2018-09-13 10 48 36

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -201,13 +201,31 @@ class Loader extends React.Component<Props, State> {
   }
 
   static getDerivedStateFromProps(props: Props, state: State): State {
-    const routeInformation = getRouteInformation(props) || {};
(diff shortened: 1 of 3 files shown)


opyh:
@lennerd Yeah, a very simple design for the desktop version would be cool to have as part if this PR. This still looks too broken to show it to someone.

dakeyras7:
Code lgtm 👍🏼

lennerd:
What do you think?

screenshot 2018-09-13 13 40 31

opyh:
I think it's great! :)

PR #114, opened 2018-09-13
Restore Last Selected Category When Closing Node Toolbar

dakeyras7:
How things were before the fix:

* Generally the selected category is fetched from the URL and passed down the component tree
* Except for when the URL contains a feature id (i.e. the NodeToolbar is open). Then we keep the previously selected category in the app state and pass that down the component tree.
* If then the NodeToolbar is closed again, we don't properly restore the previously selected category into the URL again. We instead change the URL to '/beta', so that any subsequent renders, will eventually clear out any previously selected category

After the fix:

The fix now restores the correct URL containing a previously selected category when the NodeToolbar is closed.

### DIFF
--- a/src/components/CloseLink.js
+++ b/src/components/CloseLink.js
@@ -24,13 +24,13 @@ type Props = {
 
 class CloseLink extends React.Component<Props> {
   onClick = event => {
+    event.preventDefault();
+    event.stopPropagation();
+
     if (this.props.onClick) {
       this.props.onClick(event);
       return;
     }
-
-    event.preventDefault();
-    event.stopPropagation();
 
     const params = getQueryParams();
     this.props.history.push({ pathname: '/beta', search: queryString.stringify(params) });
--- a/src/App.js
+++ b/src/App.js
@@ -202,6 +202,7 @@ class Loader extends React.Component<Props, State> {
 
   static getDerivedStateFromProps(props: Props, state: State): State {
     const routeInformation = getRouteInformation(props);
+
     let result: $Shape<State> = {
       equipmentInfoId: null,
       category: null,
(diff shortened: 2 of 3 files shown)


src/components/CloseLink.js
```diff
@@ -24,14 +24,14 @@ type Props = {
 
 class CloseLink extends React.Component<Props> {
   onClick = event => {
+    event.preventDefault();
+    event.stopPropagation();
```
lennerd: Should this be handled by the component in every case? Maybe a onClick prop is given to overwrite exactly this. No?

dakeyras7: Yeah, maybe you are right @lennerd. I didn't write it just moved it up, so the logic works.

dakeyras7:
Oh, I see. This is what I supspected though. Manually writing the URL will become tedious long-term. Let me see if I can add it for "correctness" sake.

dakeyras7:
I'm just realising that my fix also breaks the previous behavior, where the back button on the "I have a problem" dialog completely closes the node Toolbar now. I have a suspicion what I did wrong here. Let me fix these things.

dakeyras7:
I fixed a couple of issues now:

* In my previous attempt clicking 'back' on a subpage of the node toolbar like (ReportDialog) would close the whole node toolbar. This is now fixed and closing a subpage navigates back to the node toolbar.
* In several situations like opening the wheelchair a11y editor, toilet a11y editor or report dialog the current filters (query params in the URL) would just be cleared. This is now fixed and they stay until the user decides to clear them intentionally.

Pls review and verify

mutaphysis:
Rebased against master, still need verification

PR #116, opened 2018-09-13
Fix jumping map.

lennerd:
This PR fixed a jumping map when closing the "new place" modal by storing the map viewport state in the app state.

Fixes: https://trello.com/c/Y7cqtE9f

### DIFF
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -109,11 +109,7 @@ export default class Map extends React.Component<Props, State> {
     const map = this.map;
     if (!map) return;
     const { lat, lng } = map.getCenter();
-    const zoom = Math.max(
-      map.getZoom(),
-      this.props.minZoomWithSetCategory,
-      this.props.minZoomWithoutSetCategory
-    );
+    const zoom = map.getZoom();
 
     const onMoveEnd = this.props.onMoveEnd;
     if (typeof onMoveEnd === 'function') {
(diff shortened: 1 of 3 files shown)


src/components/Map/Map.js
```diff
@@ -541,23 +542,36 @@ export default class Map extends React.Component<Props, State> {
...
 
-    // use the feature coordinates
     const feature = props.feature;
+
+    let center;
+
+    // Use the feature coordinates
     if (
       feature &&
+      getFeatureId(feature) !== this.lastFeatureId &&
```
mutaphysis: Could this not lead to weird behavior if the feature id did not change? Or are the else ifs not causing any side effects when this is called without a feature change?

mutaphysis: Not sure if there even is a way to trigger this.

lennerd: I'm not 100% sure what you mean. 😕 

This line guarantees the center is only set to the features coordinates when the marker/feature was first selected. Any future changes in the tree will use the app state center position until a new marker was selected (or the old one simply deselected).

lennerd: The app state center position is updated everytime the map was updated (moveend, zoomend). Also when the marker was clicked and moved into the map center.

mutaphysis: The behavior in the `else` branches changed. I'll better just come over :D

mutaphysis: Makes sense!

PR #117, opened 2018-09-13
Use photon results for category icon in search results.

lennerd:
Fixes: https://trello.com/c/Y7cqtE9f

Not sure about the `Icon` component. Looks like its used only for displaying category icons. Should we rename it to `CategoryIcon`?

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -2,8 +2,7 @@
 
 import styled from 'styled-components';
 import * as React from 'react';
-import type { NodeProperties, YesNoLimitedUnknown } from '../lib/Feature';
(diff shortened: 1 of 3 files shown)


src/components/Icon.js
```diff
@@ -87,9 +85,9 @@ const StyledIconContainer = styled('figure')`
   }
 `;
 
+// @TODO Rename it to CategoryIcon
 export default function Icon({
```
mutaphysis: Good Idea

lennerd: Should the renaming be part of this PR?

mutaphysis: Can be done separately

PR #118, opened 2018-09-14
WIP: Fix compiler warnings

lennerd:
This PR might need very thorough reviews as it changes the usage of state and props dramatically due to the react/no-deprecated warning deprecating [`componentWillReceiveProps`](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops).

The difficult part (at least for me learning something new about the project structure every day) is the how `componentWillReceiveProps` is used in most of the cases. While used for fetching data it also often changes the state before the component is rendered without triggering a loop of rerendering the component. As we cannot simply change the state in `componentDidUpdate` without risking a rerender loop one need to combine it with `getDerivedStateFromProps`. I think pair programming (with @opyh or @mutaphysis) might help to get a safe solution more quickly.

Affected Components:

* [x] RadioStatusEditor
* [ ] Map
* [ ] AccessibilitySourceDisclaimer
* [ ] BreadCrumbs
* [ ] EquipmentOverview
* [ ] ShareButton
* [ ] LicenseHint
* [ ] NodeToolbarFeatureLoader
* [ ] PhotoNotifcation
* [ ] PhotoSection
* [ ] FixOnExternalPage
* [ ] ReportDialog
* [ ] SourceLink
* [ ] PhotoUploadCaptchaToolbar
* [ ] SearchResult
* [ ] Toolbar

### DIFF
--- a/src/components/CreatePlaceDialog/CreatePlaceDialog.js
+++ b/src/components/CreatePlaceDialog/CreatePlaceDialog.js
@@ -150,7 +150,7 @@ export default class CreatePlaceDialog extends React.Component<Props> {
   renderAppLinkListElements() {
     return this.appLinks().map(link => (
       <li key={link.title}>
-        <a className="link-button" href={link.href} target="_blank">
+        <a className="link-button" href={link.href} target="_blank" rel="noopener noreferrer">
           <AppIcon src={link.icon} alt="" aria-hidden />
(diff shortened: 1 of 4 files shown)


mutaphysis:
We can do more PRs for the other features later, merged for now

PR #120, opened 2018-09-16
Focus Management Improvements

dakeyras7:
This fixes a and cleans up a lot of things.

* wicg-focus-ring library is replaced with the newer more standards conform focus-visible. thus all occurences of `focus-ring` are now replaced with `focus-visible`.
* We use a `:focus-within` polyfill now so that we can use it cross-browser to give a `RadioGroup` a focus ring when one of its `RadioButton`s is in focus for example.
* The `postcss-focus-within` plugin duplicates any `.focus-within` class added by the polyfill into a `:focus-within` rule.
* For focus trapping we are using the `focus-trap` library and its accompanying convenience wrapper for React `focus-trap-react`. Because some items don't work well with `focus-trap-react`s additionally added `<div>` tag, I added a custom `withFocusTrap` HOC that works similar but without any `<div>` tag.
* Finally I searched through the code and tried to remove all occurences of custom focus management and use the new `FocusTrap` and `withFocusTrap`.


src/index.css
```diff
@@ -47,22 +47,30 @@ body:before {
...
 
-*.focus-ring {
+/*
+  We use a stronger and consistent focus indicator when an element receives focus via
+  keyboard.
+*/
+.js-focus-visible .focus-visible {
+  outline: none;
   border-radius: 4px;
   box-shadow: inset 0px 0px 0px 2px #4469E1;
```
opyh: Could you move the #4469E1 color string into `Colors.js`?

package.json
```diff
@@ -19,6 +19,10 @@
     "es6-symbol": "^3.1.1",
     "esri-leaflet": "^2.1.4",
     "fast-levenshtein": "^2.0.6",
+    "focus-trap": "^3.0.0",
+    "focus-trap-react": "https://github.com/sozialhelden/focus-trap-react.git",
```
opyh: Could you add a git commit reference here to be safer?

src/index.css
dakeyras7: This is a css file and we can use the global colors from `Colors.js` here once we are using the globals mechanism of StyledComponents probably.

opyh: Gna, of course :) Then in the merged other version.

PR #122, opened 2018-10-10
Allow switching back to place details from equipment

mutaphysis:
- added base button component
- updated flow types (unrelated but fixed annoying styled components error)
- use existing routes for handling the switch back to place details


src/components/Button.js
```diff
@@ -0,0 +1,50 @@
...
+  render() {
+    return (
+      <button
+        ref={button => (this.button = button)}
+        className={this.props.className || ''}
+        onBlur={this.props.onBlur}
+        onFocus={this.props.onFocus}
+        onClick={this.props.onClick}
+        onKeyDown={this.props.onKeyDown}
+        aria-label={this.props.ariaLabel}
```
lennerd: Shouldn`t this be named `ariaLabel`? And why not simply passing down all props (including children) with `<button {...props} />`?

src/components/Button.js
```diff
@@ -0,0 +1,50 @@
...
+  className: string,
+  ariaLabel: ?string,
+  onClick: ClickHandler,
+  onFocus?: FocusHandler,
+  onBlur?: BlurHandler,
+  onKeyDown?: KeyDownHandler,
+  children?: React$Node,
+};
+
+class Button extends React.Component<Props> {
```
lennerd: Isn't this a common component? Maybe this can be moved into `/components/common`.

src/components/NodeToolbar/NodeToolbarFeatureLoader.js
```diff
@@ -21,14 +21,14 @@ type Props = {
...
   modalNodeState: ModalNodeState,
-  isReportMode: boolean,
   onOpenReportMode: ?() => void,
   onStartPhotoUploadFlow: () => void,
   history: RouterHistory,
   onClose?: ?() => void,
   onClickCurrentMarkerIcon?: (feature: Feature) => void,
   onSelectWheelchairAccessibility?: (newValue: YesNoLimitedUnknown) => void,
   onEquipmentSelected: (placeInfoId: string, equipmentInfo: EquipmentInfo) => void,
+  onShowPlaceDetails: (featureId: string | number) => void,
```
lennerd: The `NodeToolbarFeatureLoader` is not part of this PR, I know. I still have the feeling this helper component can be greatly simplified. By creating the promise outside the component and passing it via props you can remove alot of code checking if the passed promise value is the same or not. I just created a similar example for the async header. Come by if you would like to further discuss this. But as I said: not part of this PR.

src/components/Button.js
mutaphysis: I just copied CloseLink, I have to admit :)

src/components/Button.js
mutaphysis: Just copied CloseLink and put it next to it.

src/components/NodeToolbar/NodeToolbarFeatureLoader.js
mutaphysis: If there is some reusable code for the AsyncHeader, feel free to create the PR :)

src/components/Button.js
mutaphysis: The /common pattern does not exist yet, can be a separate change.

lennerd: You're right! Than this should be fine I guess. 👍

src/components/Button.js
mutaphysis: `aria-label` is aligned with how react treats these - see https://reactjs.org/docs/accessibility.html

lennerd: Again, you are right. Just strange they do not conform to the rest of the props they use (`className`, `htmlFor` etc.). Anyway, passing down all props could make this component more reusable I guess. But we can leave it as it is for now if you think this is not part of this PR.

mutaphysis:
You can also review individual comments, so the flow types are not visible.

src/components/Button.js
lennerd: -1 for not passing down all props.

src/components/Button.js
lennerd: 0 for the component not being part of a common folder.

src/components/NodeToolbar/NodeToolbarFeatureLoader.js
lennerd: 0 for generating the promise inside the component. But as we said, this should be something we can use as a common component.

lennerd:
Like this! Added the weights. Will definitely use it for the next PR!

src/components/Button.js
mutaphysis: It seems passing all props to an html component is considered an anti pattern - https://vasanthk.gitbooks.io/react-bits/anti-patterns/07.spreading-props-dom.html

Also I cannot safely type this with flow, typescript at least has prop types :/

PR #123, opened 2018-10-10
Fixed wheelmap base url for proxy usage

mutaphysis:
- Also ensure that base url is actually passed into map


src/lib/cache/WheelmapFeaturePhotosCache.js
```diff
@@ -6,10 +6,12 @@ import env from '../env';
 
 export default class WheelmapFeaturePhotosCache extends URLDataCache<?WheelmapFeaturePhotos> {
   getPhotosForFeature(featureId: string): Promise<?WheelmapFeaturePhotos> {
+    const wheelmapApiBaseUrl = env.public.wheelmap.baseUrl
```
lennerd: 0 — Can be simplified to `env.public.wheelmap.baseUrl
      || env.public.baseUrl || '/'`

PR #125, opened 2018-10-11
Use lightweight wheelmap feature before loading full version

mutaphysis:


src/components/NodeToolbar/NodeToolbarFeatureLoader.js
```diff
@@ -29,6 +30,27 @@ type Props = {
...
+  onOpenToiletAccessibility: () => void,
+  onOpenWheelchairAccessibility: () => void,
+  onCloseWheelchairAccessibility: () => void,
+  onCloseToiletAccessibility: () => void,
+  onClickCurrentMarkerIcon?: (feature: Feature) => void,
+  onEquipmentSelected: (placeInfoId: string, equipmentInfo: EquipmentInfo) => void,
+  onShowPlaceDetails: (featureId: string | number) => void,
+  // Simple 3-button wheelchair status editor
+  presetStatus: YesNoLimitedUnknown,
+  onSelectWheelchairAccessibility: (value: YesNoLimitedUnknown) => void,
```
lennerd: 0: What about having this extra container for the node toolbar we talked about so that we do not need to pass down dozens of event handlers?

lennerd:
@opyh just explained the difference between them. Thanks! :)

PR #126, opened 2018-10-11
Use central category resolution also for search results

mutaphysis:
- Moved code from NodeToolbarFeatureLoader into Categories
 - Extended to support search results
 - Determine categories from search result or wheelmap result
 - Fixed a few more flow errors
 - Fixed cache usage in search results

### DIFF
--- a/src/lib/Categories.js
+++ b/src/lib/Categories.js
@@ -1,10 +1,14 @@
 // @flow
+import { t } from 'ttag';
 
 import { globalFetchManager } from './FetchManager';
-import { t } from 'ttag';
 import { translatedStringFromObject } from './i18n';
 import ResponseError from './ResponseError';
 import env from './env';
+
+import type { Feature } from './Feature';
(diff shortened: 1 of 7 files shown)


src/lib/Categories.js
```diff
@@ -85,10 +89,13 @@ export default class Categories {
...
 
   static getCategory(lookupTable: CategoryLookupTables, idOrSynonym: string | number): ACCategory {
-    if (!lookupTable.synonymCache) throw new Error('Empty synonym cache.');
+    const synonymCache = lookupTable.synonymCache;
+
+    if (!synonymCache) {
+      throw new Error('Empty synonym cache.');
+    }
 
-    // @TODO \o/ Help! Sebastian!
```
opyh: ⛑

lennerd: 👨‍⚕️ ◀️ @opyh

PR #128, opened 2018-10-11
Use NotFound as error component

lennerd:
* Remove not found and error handling from App and MainView.
* Use NotFound component in _app.

### DIFF
--- a/src/components/NotFound/NotFound.js
+++ b/src/components/NotFound/NotFound.js
@@ -5,17 +5,17 @@ import get from 'lodash/get';
 import * as React from 'react';
 import { findDOMNode } from 'react-dom';
 import styled from 'styled-components';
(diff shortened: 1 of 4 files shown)


src/components/NotFound/NotFound.js
```diff
@@ -99,7 +93,7 @@ class NotFound extends React.Component<Props> {
     return (
       <ModalDialog
         className={classList.join(' ')}
-        isVisible={this.props.isVisible}
+        isVisible={true}
```
mutaphysis: 0: hmm :)

lennerd: 😄 Should we use default props in `ModalDialog`?

mutaphysis: All is fine, nothing to see here 💃

PR #130, opened 2018-10-11
Added caching options to photo caches

mutaphysis:


src/lib/cache/EquipmentInfoCache.js
```diff
@@ -9,7 +9,10 @@ export default class EquipmentInfoCache extends FeatureCache<
   EquipmentInfo,
   EquipmentInfoFeatureCollection
 > {
-  static fetchFeature(id: string | number): Promise<Response> {
+  static fetchFeature(
+    id: string | number,
+    options: { useCache: boolean } = { useCache: true }
```
lennerd: -1: `useCache` is not passed to the fetch method?

PR #131, opened 2018-10-11
Add redirects for homepage (/map) and legacy urls.

lennerd:
Redirects to classic.wheelmap.org:

* `/community_support/new`
* `/map/japan/post_office/wheelchair/yes`
* `/embed`

Redirects `/map` to homepage (`/`).
Routes for `/search` and `/nodes` match the current path. 

@opyh @mutaphysis what about `/en/nodes`, etc.?

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -12,8 +12,13 @@ app.prepare().then(() => {
   const server = express();
 
   // Old urls from the classic rails app
+  // First capturing group represents optional locale
   server.get(
-    [/([a-zA-Z_-]*)\/map\/.+/, /([a-zA-Z_-]*)\/community_support\/new.*/, /([a-zA-Z_-]*)\/embed.*/],
+    [
+      /(\/[a-zA-Z_-]+)?\/map\/.+/,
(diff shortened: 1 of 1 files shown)


src/server.js
```diff
@@ -11,6 +11,20 @@ const handle = app.getRequestHandler();
 app.prepare().then(() => {
   const server = express();
 
+  // Old urls from the classic rails app
+  server.get(
+    [/([a-zA-Z_-]*)\/map\/.+/, /([a-zA-Z_-]*)\/community_support\/new.*/, /([a-zA-Z_-]*)\/embed.*/],
```
mutaphysis: -1: Shouldn't all `[a-zA-Z_-]*` be `[a-zA-Z_-]+`?

lennerd: This is part of the regexp represents the locale, which is optional.

mutaphysis: I guess I am just nit-picky about undocumented regexp. But the extra slash is not optional - so it only matches eg. `/community_support/new` but not `community_support/new`. Either the whole first part is optional, then the slash should be optional. Or it is required, then it should need at least one character before the slash.

PR #140, opened 2018-10-15
Make SearchInputField uncontrolled and debouncing its value.

lennerd:
This does not show the loading indicator when someone is typing but rather debouncing the whole onChange handler of the SearchInputField whenever a new value was entered.

**This also means the `SearchInputField` is uncontrolled and will not response to changes in the URL once mounted.** I tried another way of storing a timestamp in the state of the input component to make sure changes coming from outside are distinguishable from `setState` changes inside the component. But that felt really hacky and at same time makes it hard to understand what is happening in `SearchInputField` component excatly.

Looking forward to hear your thoughts on this.

### DIFF
--- a/src/components/SearchToolbar/SearchInputField.js
+++ b/src/components/SearchToolbar/SearchInputField.js
@@ -3,17 +3,18 @@
 import * as React from 'react';
 import styled from 'styled-components';
 import { interpolateLab } from 'd3-interpolate';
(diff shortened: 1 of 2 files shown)


lennerd:
Due to changes on how the route name and the search bar expanded status will be connected with my next PR this uncontrolled search input field will show invalid values when switching between routes.

PR #142, opened 2018-10-15
Open modal dialog when using inline place accessibility editing

mutaphysis:
- renamed presetStatus to accessibilityPresetStatus
- removed old url & query handling for the inline button row
- fixed some minor flow errors

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -79,6 +79,7 @@ type State = {
   isOnboardingVisible: boolean,
   isMainMenuOpen: boolean,
   modalNodeState: ModalNodeState,
+  accessibilityPresetStatus?: ?YesNoLimitedUnknown,
   isSearchBarVisible: boolean,
   isOnSmallViewport: boolean,
   isSearchToolbarExpanded: boolean,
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -563,9 +564,9 @@ class Loader extends React.Component<Props, State> {
 
   onSelectWheelchairAccessibility = (value: YesNoLimitedUnknown) => {
     if (this.props.featureId) {
-      this.props.history.push({
-        pathname: `/beta/nodes/${this.props.featureId}/edit-wheelchair-accessibility`,
-        search: `presetStatus=${value}`,
+      this.setState({
+        modalNodeState: 'edit-wheelchair-accessibility',
```
lennerd: -1: Should the `modalNodeState` be part of the actual route so we do not have two state machines but one based on routing?

mutaphysis: Last discussion status with @opyh & you on Thursday was that we wanted to remove this from the routes. I guess we can bring all of that back, but its unrelated to the fix, the removed lines don't even work.

lennerd: Really? Than I changed my mind. Anyway, you're right. This is probably not part of this PR.

mutaphysis: I am all in for fewer state machines!

lennerd: Should we add a ticket in Trello?

PR #144, opened 2018-10-15
Allow snapping back to feature from node toolbar

mutaphysis:


src/components/Map/Map.js
```diff
@@ -744,6 +744,21 @@ export default class Map extends React.Component<Props, State> {
     if (this.mapElement) this.mapElement.focus();
   }
 
+  snapToFeature() {
```
lennerd: Nice idea to handle imperative map state!

PR #145, opened 2018-10-15
Fix search toolbar and button

lennerd:
This PR tries to fix https://trello.com/c/GEzQ1tBS and https://trello.com/c/IQ6cw1Kj

Especially the state of the App is tricky and needs refactoring IMHO. Do not feel very confident. Please check if you see any issues.

About this PR:

* Debounce search requests via debouncing promise in SearchData
* Tries to fix Webpack warning of invalid import to also fix HMR (warning still shown and HMR still not working though)
* Fix state to show search button again (when NodeToolbar is shown, the map is not focused, etc.)


src/lib/debouncePromise.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import debounce from 'lodash/debounce';
+
+function debouncePromise<T>(
+  fn: (...args: any[]) => Promise<T>,
+  delay: number
+): (...args: any[]) => Promise<T> {
+  let promise;
```
mutaphysis: -1: Name is really bad, this is never a promise. Also not too happy with the order of calls/assignments, maybe you can move the promiseHolder `promise` down, and pass it as an argument into the debounce function? Would have been easier to read for me. Typing for this would also be nice.

src/app/SearchData.js
```diff
@@ -66,17 +67,19 @@ const SearchData: DataTableEntry<SearchProps> = {
 
     const searchQuery = query.q;
     let trimmedSearchQuery;
-    let searchResults = Promise.resolve({ features: [] });
+    let searchResults: Promise<SearchResultCollection> | SearchResultCollection = {
+      features: [],
+    };
 
     if (searchQuery && (trimmedSearchQuery = searchQuery.trim())) {
```
mutaphysis: 0: Why not move trimming into if? I can't see where it is used.

lennerd: The trimmed search query is used when creating the request in line 75. It simply checks if the string is not empty for which we do not need to make a request to Photon.

src/lib/debouncePromise.js
lennerd: The function debounces (verb) a promise by a given delay. Thats why I named it like this. What do you think would fit better?

Good idea of passing the promise into the debounce function. Will fix that.

lennerd: Passing the promise into the debounce function does not work as we need to set the promise holder `promise` to null once the debounce function is finally called. 😕 

The promise holder is also automatically typed by flow due to its following usage inside the function.

screenshot 2018-10-15 17 15 12

mutaphysis: I was talking about the local variable `promise`, which is not a promise.

src/app/SearchData.js
mutaphysis: I don't the reason for a local variable with `let` then.

src/lib/debouncePromise.js
lennerd: Ahh … sorry. Now I see. What do you think about `promiseHolder` or `promiseCache?

src/app/SearchData.js
mutaphysis: I meant 'used elsewhere' above, sorry.

src/lib/debouncePromise.js
mutaphysis: Good point - I did not think about the resetting. Either name is fine :)

src/app/SearchData.js
lennerd: This would make the code more complex IMO due to more nesting.

``` js
if (searchQuery) {
  let trimmedSearchQuery = searchQuery.trim()

  if (trimmedSearchQuery) {
    searchResults = (isServer ? searchPlaces : searchPlacesDebounced)(trimmedSearchQuery, {
      lat: parseFloat(query.lat),
      lon: parseFloat(query.lon),
    });

    // Fetch search results when on server. Otherwise pass (nested) promises as props into app.
    searchResults = isServer ? await searchResults : searchResults;
  }
}
```

Compared to this:

``` js
let trimmedSearchQuery;

if (searchQuery && (trimmedSearchQuery = searchQuery.trim())) {
  searchResults = (isServer ? searchPlaces : searchPlacesDebounced)(trimmedSearchQuery, {
    lat: parseFloat(query.lat),
    lon: parseFloat(query.lon),
  });

  // Fetch search results when on server. Otherwise pass (nested) promises as props into app.
  searchResults = isServer ? await searchResults : searchResults;
}
```

opyh: Why no early returns?

lennerd: Good point! PR is already merged. But I will maybe add it to another PR (… living on the edge and so forth).

PR #149, opened 2018-10-16
Added cordova builds

mutaphysis:
- inject build time properties (such as categories)
- refresh built-in properties and cache in local storage
- ios building & routing working

Open issues:

- Sometimes navigation to gap://ready happens, timing issue with deviceReady.
- android untested


src/components/PhotoUpload/PhotoUploadInstructionsToolbar.js
```diff
@@ -122,9 +122,9 @@ const StyledToolbar = styled(Toolbar)`
               background-color: #EEE;
             }
 
-            .entrance-image { background-image: url('/static/images/photo-upload/entrancePlaceholder.png'); }
-            .sitemap-image { background-image: url('/static/images/photo-upload/sitemapPlaceholder.png'); }
-            .toilet-image { background-image: url('/static/images/photo-upload/toiletPlaceholder.png'); }
+            .entrance-image { background-image: url('static/images/photo-upload/entrancePlaceholder.png'); }
+            .sitemap-image { background-image: url('static/images/photo-upload/sitemapPlaceholder.png'); }
+            .toilet-image { background-image: url('static/images/photo-upload/toiletPlaceholder.png'); }
```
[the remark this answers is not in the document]
mutaphysis: Ios cordova is not running under file:/// but som arbitrary path

src/app/getInitialProps.js
```diff
@@ -115,28 +116,32 @@ export async function getAppInitialProps(
...
   const userAgentParser = new UAParser(userAgentString);
   const userAgent = ((userAgentParser.getResult(): any): UAResult);
 
-  const clientSideConfiguration = clientCache.clientSideConfiguration
-    ? clientCache.clientSideConfiguration
-    : await fetchClientSideConfiguration(hostName);
+  const clientSideConfiguration =
+    useCache && clientCache.clientSideConfiguration
+      ? clientCache.clientSideConfiguration
+      : await fetchClientSideConfiguration(hostName);
```
opyh: 1) This way the promises are awaited serially, not in parallel, which might make loading quite slow. `await Promise.all()` would be cool!
2) How about removing the code duplication here:
```javascript
const dataFetchers = {
  clientSideConfiguration: async () => await fetchClientSideConfiguration(hostName),
  translations: () => loadExistingLocalizationByPreference(locales),
  categories: async () => await Categories.generateLookupTables({ locale: preferredLocale });
}

const cachedOrFetchedResults = Object.keys(dataFetchers)
  .map(k => (useCache && clientCache[k]) || dataFetchers[k]())
…
return {
...await Promise.all(cachedOrFetchedResults),
…,
};
...

export function clientStoreAppInitialProps(props: $Shape<AppProps>) {
  Object.keys(dataFetchers).forEach(k => {
    clientCache[k] = props[k] || clientCache[k];
  });
}
```

src/lib/NextRouteHistory.js
```diff
@@ -26,7 +28,14 @@ class NextRouterHistory implements RouterHistory {
     const route = this.router.getRoute(name, true);
     const path = this.router.generate(name, params);
 
-    NextRouter[method]({ pathname: '/main', query: { routeName: route.name, ...params } }, path);
+    if (this.isCordova) {
+      const query = { routeName: route.name, path, ...params };
+      const pathname = window.location.pathname;
+      NextRouter[method]({ pathname: '/cordova', query }, { pathname, query });
```
opyh: I think this solution of having two pages is okay. When we know that the implementation stays as similar as it is now after a while, we can merge the two implementations into one again – for now I'd keep both even if it means code duplication.

src/lib/config.js
```diff
@@ -1,5 +1,10 @@
 import env from './env';
 
+import isCordova from './isCordova';
+
+// use base url on server or in cordova builds, otherwise use the proxy running on the hosting server
+const useWheelmapBaseUrl = typeof window === 'undefined' || isCordova();
```
opyh: How about `useAbsoluteWheelmapBaseUrl`?

src/lib/savedState.js
```diff
@@ -3,6 +3,7 @@
 import storage from 'local-storage-fallback';
 
 const lastMoveDateString = storage.getItem('wheelmap.map.lastMoveDate');
+const initialPropsString = storage.getItem('wheelmap.initialProps');
```
opyh: Nice idea :)

src/pages/_document.js
```diff
@@ -35,7 +38,8 @@ export default class MyDocument extends Document {
             content={`
               default-src
                 ws:
-                gap:
+                ${isCordovaBuild ? cordovaBlock : ''}
+                file:
```
opyh: The line with `file:` shouldn't be here

src/pages/cordova.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    this.state.isDeviceReady = isDeviceReady;
+  }
+
+  componentDidMount() {
+    const userAgentString = window.navigator.userAgent;
+    const languages = [window.navigator.language]
+      .concat(window.navigator.languages || [])
+      .filter(Boolean);
```
opyh: Did you test language fallback here?

src/pages/cordova.js
```diff
@@ -0,0 +1,149 @@
...
+} from '../lib/i18n';
+import isCordova from '../lib/isCordova';
+
+// dynamically load app only after cordova is ready
+const DynamicApp = dynamic(import('../App'), {
+  ssr: false,
+});
+
+// handle ready before mounting the CordovaMain page
+let isDeviceReady = false;
```
opyh: Where is this variable used? Is it used? :)

src/lib/config.js
mutaphysis: Done

src/app/getInitialProps.js
mutaphysis: Good catch will add another PR for this.

src/pages/_document.js
mutaphysis: Done

PR #150, opened 2018-10-17
Add link component. Add an example on how to use it in category filters.

lennerd:
This PR contains:

* A new Link component using the React Context API to pass around the route history, current params and current route name
* Refactored buttons as an example of the new link component used for the category search toolbar filters
* Simplifications for the router history implementations by adding a base class 

Example for the link component:

```js
<RouteConsumer>
  {context => {
    const params = { ...context.params };

    if (showCloseButton) {
      delete params.category;
    } else {
      params.category = category;
    }

    return (
      <Link
        params={params}
        routeName={context.name}
        aria-label={showCloseButton ? t`Remove ${props.name} Filter` : props.name}
        className={className}
        onFocus={props.onFocus}
      >
        <IconButton
          isHorizontal={showCloseButton}
          caption={props.name}
          hasCircle={props.hasCircle}
        >
          {icon}
        </IconButton>
        {showCloseButton && <CloseIcon />}
      </Link>
    );
  }}
</RouteConsumer>
```

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,7 +9,7 @@ import config from './lib/config';
 import savedState, { saveState, isFirstStart } from './lib/savedState';
 import { hasBigViewport, isOnSmallViewport } from './lib/ViewportSize';
 import { isTouchDevice, configureUserAgent, type UAResult } from './lib/userAgent';
-import { type RouterHistory as NewRouterHistory } from './lib/RouterHistory';
+import { type RouterHistory } from './lib/RouterHistory';
 import { type SearchResultCollection } from './lib/searchPlaces';
 import type { WheelmapFeature } from './lib/Feature';
 import type { SearchResultFeature } from './lib/searchPlaces';
@@ -37,6 +37,7 @@ import { type CategoryLookupTables } from './lib/Categories';
 import { type PhotoModel } from './lib/PhotoModel';
 import { type PlaceDetailsProps } from './app/PlaceDetailsProps';
 import { type PlaceFilter } from './components/SearchToolbar/AccessibilityFilterModel';
+import { RouteProvider } from './components/Link';
 
 import 'react-activity/dist/react-activity.css';
 import './App.css';
@@ -52,7 +53,7 @@ export type Link = {
 
 type Props = {
   className?: string,
-  routerHistory: NewRouterHistory,
+  routerHistory: RouterHistory,
   routeName: string,
   categories?: CategoryLookupTables,
   userAgent?: UAResult,
@@ -299,24 +300,6 @@ class App extends React.Component<Props, State> {
     }
 
     this.props.routerHistory.push(routeName, params);
-  };
-
-  onCategorySelect = (category: string) => {
-    const { routeName, routerHistory } = this.props;
-    const params = this.getCurrentParams();
-
-    params.category = category;
-
-    routerHistory.push(routeName, params);
-  };
-
-  onCategoryReset = () => {
-    const { routeName, routerHistory } = this.props;
-    const params = this.getCurrentParams();
-
-    delete params.category;
-
-    routerHistory.push(routeName, params);
   };
 
   onClickFullscreenBackdrop = () => {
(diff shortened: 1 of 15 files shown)


src/App.js
```diff
@@ -9,7 +9,7 @@ import config from './lib/config';
 import savedState, { saveState, isFirstStart } from './lib/savedState';
 import { hasBigViewport, isOnSmallViewport } from './lib/ViewportSize';
 import { isTouchDevice, configureUserAgent, type UAResult } from './lib/userAgent';
-import { type RouterHistory as NewRouterHistory } from './lib/RouterHistory';
+import { type RouterHistory } from './lib/RouterHistory';
```
mutaphysis: 🆒

src/components/Link.js
```diff
@@ -0,0 +1,85 @@
...
+
+type RouteContext = {
+  params: RouteParams,
+  history: RouterHistory,
+  name: string,
+};
+
+const { Provider: RouteProvider, Consumer: BaseRouteConsumer } = createContext<?RouteContext>();
+
+const RouteConsumer = ({ children }: { children: (context: RouteContext) => React$Node }) => (
```
mutaphysis: 0: Maybe extract to its own file

src/components/Link.js
```diff
@@ -0,0 +1,85 @@
...
+
+class Link extends Component<Props> {
+  handleClick(
+    history: RouterHistory,
+    params?: RouteParams,
+    event: SyntheticEvent<HTMLLinkElement>
+  ) {
+    const { routeName, replace, onClick } = this.props;
+
+    onClick && onClick(event);
```
mutaphysis: 0: use real if?

src/components/Link.js
```diff
@@ -0,0 +1,85 @@
...
+    const { routeName, params, ...props } = this.props;
+    const href = history.generate(routeName, params);
+
+    delete props.replace;
+
+    return (
+      <a
+        {...props}
+        href={href}
+        onClick={event => {
```
mutaphysis: 0: use real arrow-function property instead of inline to prevent reallocation on each render?

src/components/SearchToolbar/CategoryButton.js
```diff
@@ -44,24 +45,40 @@ function CategoryButton(props: Props) {
...
-      aria-label={showCloseButton ? t`Remove ${props.name} Filter` : props.name}
-    >
-      <IconButton isHorizontal={showCloseButton} caption={props.name} hasCircle={props.hasCircle}>
-        {icon}
-      </IconButton>
-      {showCloseButton && <CloseIcon />}
-    </Button>
+    <RouteConsumer>
+      {context => {
+        const params = { ...context.params };
```
mutaphysis: 0: Maybe do this, as the Category gets replaced anyway

```suggestion
        const { prevCategory, ...params } = context.params;
```

src/components/SearchToolbar/CategoryButton.js
```diff
@@ -44,24 +45,40 @@ function CategoryButton(props: Props) {
...
+
+        if (showCloseButton) {
+          delete params.category;
+        } else {
+          params.category = category;
```
mutaphysis: 0: Totally unrelated, but it's confusing that `this.props.category` is not a Category but a categoryId

src/lib/RouterHistory.js
```diff
@@ -1,7 +1,43 @@
...
+export type RouterHistoryMethod = 'push' | 'replace';
+
+export type RouteParams = { [name: string]: any };
 
 export interface RouterHistory {
-  push(name: string, params?: { [name: string]: any }): void;
+  push(name: string, params?: RouteParams): void;
+  replace(name: string, params?: RouteParams): void;
+  getRoute(name: string): Route;
+  generate(name: string, params?: RouteParams): string;
```
opyh: if this is generating a path name, I'd prefer `generatePathName` or `generatePath`

lennerd: Its generating a path name including query parameters. Will fix.

src/components/Link.js
lennerd: The alternative would be to use the `.bind` function, which also creates a new function when called. 😕

src/components/SearchToolbar/CategoryButton.js
lennerd: This is not a property but a param. But you're right! Would fix this in a separate task/PR.

src/App.js
```diff
@@ -649,48 +643,54 @@ class App extends React.Component<Props, State> {
...
-        onFinishReportPhotoFlow={this.onFinishReportPhotoFlow}
-        onAbortReportPhotoFlow={this.onExitReportPhotoFlow}
-      />
+      >
+        <MainView
+          {...extraProps}
+          ref={mainView => {
+            this.mainView = mainView;
+          }}
+          onClickSearchButton={this.onClickSearchButton}
```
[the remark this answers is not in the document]
lennerd: Yes, please! 😍

src/components/SearchToolbar/CategoryButton.js
lennerd: Personally I like the descriptive delete more than extracting an object property without using it. I'm always worried somebody working with the code thinks its a mistake and removes the property from the object spreading. What do you think?

lennerd:
Added changes based on your feedback @opyh and @mutaphysis and rebased this PR on feat/ssr.

src/components/SearchToolbar/CategoryButton.js
mutaphysis: Fine by me :)

src/components/Link.js
mutaphysis: Ah, I see, you would lose the routeContext, sorry missed that!

PR #151, opened 2018-10-17
Add open graph and twitter share images.

lennerd:


src/app/PlaceDetailsData.js
```diff
@@ -130,7 +130,9 @@ const PlaceDetailsData: DataTableEntry<PlaceDetailsProps> = {
 
       const equipmentPromise = equipmentInfoId ? fetchEquipment(equipmentInfoId, useCache) : null;
       const equipmentInfo = equipmentPromise
-        ? isServer ? await equipmentPromise : equipmentPromise
```
opyh: I think nested ternary operators are a bit difficult to understand – how about this? :)

```javascript
(isServer ? await equipmentPromise : equipmentPromise) || null
```

src/app/PlaceDetailsData.js
```diff
@@ -223,48 +226,65 @@ const PlaceDetailsData: DataTableEntry<PlaceDetailsProps> = {
...
               property="place:location:latitude"
               key="place:location:latitude"
-            />
-          ),
-          coordinates != null && (
+            />,
             <meta
               content={coordinates[0]}
               property="place:location:longitude"
               key="place:location:longitud"
```
opyh: There is an `e` missing here

src/app/PlaceDetailsData.js
```diff
@@ -223,48 +226,65 @@ const PlaceDetailsData: DataTableEntry<PlaceDetailsProps> = {
         // translator: Title for sharing a place detail page
         const thisPlaceIsOn = t`This place is on ${textContent.product.name}: ${placeTitle}`;
 
-        extras = [
-          coordinates != null && (
+        if (coordinates) {
```
opyh:
```suggestion
        if (coordinates) {
          extras.push(...['longitude', 'latitude'].map((coord, i) => <meta
              content={coordinates[i]}
              property={`place:location:${coord}`}
              key={`place:location:${coord}`}
            />))
        }
```

lennerd:
@opyh Thanks for your feedback. Changes added.

PR #153, opened 2018-10-18
Add loading indicator for dynamic loaded map.

lennerd:
screenshot 2018-10-18 10 07 15


opyh:
1) Could you make the background gray so the design has more plasticity?
2) the <Dots /> could have a much bigger size

lennerd:
screenshot 2018-10-18 12 28 35

Styling fixed @opyh.

PR #154, opened 2018-10-18
Parallel loading to speed up server & client requests

mutaphysis:


src/app/PlaceDetailsData.js
```diff
@@ -120,25 +121,17 @@ const PlaceDetailsData: DataTableEntry<PlaceDetailsProps> = {
...
         ? isServer ? await equipmentPromise : equipmentPromise
         : null;
-
-      // console.log("loading", { useCache });
-      const sourcesPromise = fetchSourceWithLicense(featureId, feature, useCache);
       const sources = isServer ? await sourcesPromise : sourcesPromise;
-      // console.log("sources", sources, { useCache, feature });
-
-      const photosPromise = fetchPhotos(featureId, useCache);
       const photos = isServer ? await photosPromise : photosPromise;
```
lennerd: 0: Maybe it's cleaner to use `Promise.all`? Something like this:

```js
const [
  equipmentInfo,
  sources,
  photos
] = await Promise.all(equipmentPromise, sourcesPromise, photosPromise);
```

src/pages/_app.js
```diff
@@ -98,14 +98,16 @@ export default class App extends BaseApp {
...
+      const appPropsPromise = getAppInitialProps(
         { userAgentString, hostName, languages, ...ctx.query },
         isServer
       );
 
       if (ctx.query.routeName) {
-        routeProps = await getInitialProps(ctx.query, isServer);
+        const routePropsPromise = getInitialProps(ctx.query, isServer);
+        routeProps = await routePropsPromise;
       }
```
lennerd: 0: Similar to above. `Promise.all` might be nicer and easier to read.

PR #159, opened 2018-10-19
Fix search toolbar to be open on initial mount only on desktop devices.

lennerd:


src/App.js
```diff
@@ -491,9 +491,12 @@ class App extends React.Component<Props, State> {
   };
 
   onCloseOnboarding = () => {
+    const params = this.getCurrentParams();
+
     saveState({ onboardingCompleted: 'true' });
     this.setState({ isOnboardingVisible: false });
-    if (this.mainView) this.mainView.focusSearchToolbar();
+
+    this.props.routerHistory.push('search', params);
```
mutaphysis: How does this look in the mobile view? Will we have the full screen search popup immediately?

lennerd: 🙈 You're right! Always forgetting about mobile. Will fix it.

lennerd:
@mutaphysis thanks for the feedback! Fixed.

PR #160, opened 2018-10-19
WIP: TTL caches

lennerd:
* [x] Add `TTLCache` class
* [x] Add ClientSideConfigurationCache with TTL of an hour.
* [ ] Add TTL to CategoryCache
* [ ] Add TTL to LicenseCache and SourceCache

### DIFF
--- a/src/lib/cache/TTLCache.js
+++ b/src/lib/cache/TTLCache.js
@@ -1,0 +1,90 @@
+// @flow
+export interface TTLCacheItem<T> {
+  value: T;
+  expire: number;
+}
+
+export type TTLCacheOptions = {
+  ttl: number,
+};
+
+class TTLCache<K, V> {
+  cache: Map<K, TTLCacheItem<V>> = new Map<K, TTLCacheItem<V>>();
+  options: TTLCacheOptions;
+
+  constructor(options?: TTLCacheOptions) {
+    this.options = {
+      ttl: Infinity,
+      ...options,
+    };
+  }
+
+  set(key: K, value: V, ttl: number = this.options.ttl, now: number = Date.now()): boolean {
+    // Adding the value to the cache is not possible if ttl is zero.
+    if (ttl === 0) {
+      return false;
+    }
+
+    // Check for infinity in which case the item persists forever.
+    const expire = ttl < Infinity ? now + ttl : Infinity;
+
+    const item: TTLCacheItem<V> = {
+      value,
+      expire,
+    };
+
+    this.cache.set(key, item);
+
+    return true;
+  }
+
+  get(key: K, remove: boolean = false, now?: number): ?V {
+    const value = this.touch(key, now);
+
+    if (value == null) {
+      return null;
+    }
+
+    if (remove) {
+      this.remove(key);
+    }
+
+    return value;
+  }
+
+  has(key: K): boolean {
+    return this.cache.has(key);
+  }
+
+  remove(key: K): boolean {
+    return this.cache.delete(key);
+  }
+
+  touch(key: K, now: number = Date.now()): ?V {
+    const item = this.cache.get(key);
+
+    if (item == null) {
+      return null;
+    }
+
+    if (item.expire < now) {
+      this.remove(key);
+
+      return null;
+    }
+
+    return item.value;
+  }
+
+  touchAll(now?: number): void {
+    for (const key of this.cache.keys()) {
+      this.touch(key, now);
+    }
+  }
+
+  clear(): void {
+    this.cache.clear();
+  }
+}
+
+export default TTLCache;
--- a/src/lib/cache/TTLCache.js
+++ b/src/lib/cache/TTLCache.js
@@ -31,7 +31,7 @@ class TTLCache<K, V> {
 
   set(key: K, value: V, ttl: number = this.options.ttl, now: number = Date.now()): boolean {
     // Adding the value to the cache is not possible if ttl is zero.
-    if (ttl === 0) {
+    if (ttl <= 0) {
       return false;
     }
 
--- a/src/lib/cache/TTLCache.js
+++ b/src/lib/cache/TTLCache.js
@@ -6,17 +6,27 @@ export interface TTLCacheItem<T> {
 
 export type TTLCacheOptions = {
   ttl: number,
+  max: number,
 };
 
 class TTLCache<K, V> {
   cache: Map<K, TTLCacheItem<V>> = new Map<K, TTLCacheItem<V>>();
   options: TTLCacheOptions;
 
-  constructor(options?: TTLCacheOptions) {
+  constructor(options?: $Shape<TTLCacheOptions>) {
     this.options = {
       ttl: Infinity,
+      max: Infinity,
       ...options,
     };
+
+    if (this.options.ttl < 0) {
+      throw new Error('Invalid ttl.');
+    }
+
+    if (this.options.max < 1) {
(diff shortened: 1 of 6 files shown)


src/lib/cache/URLDataCache.js
```diff
@@ -1,55 +1,58 @@
...
+import TTLCache, { type TTLCacheOptions } from './TTLCache';
 
 // Provides a WhatWG-fetch-like API to make HTTP requests.
 // Caches response promises and returns an old promise if one is existing for the same URL.
 
+type URL = string;
+
 export default class URLDataCache<T> {
-  cache: { [key: string]: Promise<T> } = {};
+  cache: TTLCache<URL, Promise<T>>;
```
opyh: +1: Composition > Inheritance 👍🏻

src/lib/cache/URLDataCache.js
```diff
@@ -1,55 +1,58 @@
...
 
-  fetch(url: string, resolve: (data: T) => void, reject: (response: any) => void) {
-    this.constructor.fetch(url, { cordova: true }).then((response: Response) => {
+  constructor(options?: TTLCacheOptions) {
+    this.cache = new TTLCache<URL, Promise<T>>(options);
```
opyh: 0: This could happen directly in the declaration above

src/lib/cache/URLDataCache.js
```diff
@@ -1,55 +1,58 @@
...
-          resolve(fetchedData);
-        }, reject);
+        return this.constructor.getDataFromResponse(response);
       }
-      const error = new Error(response.statusText);
-      error.response = response;
-      return reject(response);
-    }, reject);
+
+      throw new new ResponseError(response.statusText, response)();
```
opyh:
```suggestion
      throw new ResponseError(response.statusText, response);
```

src/lib/cache/URLDataCache.js
```diff
@@ -1,55 +1,58 @@
...
 import { globalFetchManager } from '../FetchManager';
 import { t } from 'ttag';
+
 import ResponseError from '../ResponseError';
+import TTLCache, { type TTLCacheOptions } from './TTLCache';
 
 // Provides a WhatWG-fetch-like API to make HTTP requests.
 // Caches response promises and returns an old promise if one is existing for the same URL.
 
+type URL = string;
```
opyh: -1: I think this is misleading, as URL looks like a complex type on first sight, and there is a `URL` class on window in the browser. I'd leave `string` as key type – it's clear what it does and it does not imply the thought that there might be some canonical URL recognition tech behind it. Also, right in the subclass it's no URL anymore, but a hostname.

src/lib/cache/TTLCache.js
```diff
@@ -0,0 +1,90 @@
...
+  constructor(options?: TTLCacheOptions) {
+    this.options = {
+      ttl: Infinity,
+      ...options,
+    };
+  }
+
+  set(key: K, value: V, ttl: number = this.options.ttl, now: number = Date.now()): boolean {
+    // Adding the value to the cache is not possible if ttl is zero.
+    if (ttl === 0) {
```
opyh: Would use `<= 0` here to be more nice to the API user :)

src/lib/cache/TTLCache.js
```diff
@@ -0,0 +1,90 @@
...
+  options: TTLCacheOptions;
+
+  constructor(options?: TTLCacheOptions) {
+    this.options = {
+      ttl: Infinity,
+      ...options,
+    };
+  }
+
+  set(key: K, value: V, ttl: number = this.options.ttl, now: number = Date.now()): boolean {
```
opyh: +1: I like that you can use this functionally without time as a side effect.

src/lib/cache/TTLCache.js
```diff
@@ -0,0 +1,90 @@
...
+
+    this.cache.set(key, item);
+
+    return true;
+  }
+
+  get(key: K, remove: boolean = false, now?: number): ?V {
+    const value = this.touch(key, now);
+
+    if (value == null) {
```
opyh: How about `===`?

src/lib/cache/TTLCache.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  remove(key: K): boolean {
+    return this.cache.delete(key);
+  }
+
+  touch(key: K, now: number = Date.now()): ?V {
+    const item = this.cache.get(key);
+
+    if (item == null) {
```
opyh: How about `===`?

src/lib/cache/TTLCache.js
```diff
@@ -0,0 +1,90 @@
...
+      this.remove(key);
+
+      return null;
+    }
+
+    return item.value;
+  }
+
+  touchAll(now?: number): void {
+    for (const key of this.cache.keys()) {
```
opyh: How about this?

```suggestion
this.cache.keys().forEach((key) =>  this.touch(key, now));
```

…as this does not use the regenerator runtime

src/lib/cache/URLDataCache.js
lennerd: How do you pass options into the cache then?

src/lib/cache/URLDataCache.js
lennerd: Good catch!

src/lib/cache/TTLCache.js
lennerd: As explained in slack, this also tests against both `null` and `undefined`.

src/lib/cache/TTLCache.js
lennerd: I thought the regenerator runtime is only included when used with Generator or async functions? The generator class itself is not needing the polyfill AFAIK. Will investigate.

Also the regenerator runtime is already included because of the async functions when fetching intial props.

opyh: https://github.com/airbnb/javascript/issues/1271 has a long discussion of many aspects of this if you're interested :)

src/lib/cache/TTLCache.js
opyh: Ah – if this was the intention, a small comment would be cool :)

src/lib/cache/TTLCache.js
opyh: (Also, you seem to be right, and to have that for loop there is totally okay.)

src/lib/cache/TTLCache.js
```diff
@@ -33,11 +43,23 @@ class TTLCache<K, V> {
...
+      this.shift();
+    }
+
     this.cache.set(key, item);
 
     return true;
   }
 
+  shift() {
+    const key = this.cache.keys().next().value;
```
opyh: I'm not sure – do have ES6 maps have a specified order? If not, this could remove the last inserted item!

lennerd: Yes, they have: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared

> The keys in Map are ordered while keys added to object are not. Thus, when iterating over it, a Map object returns keys in order of insertion.

opyh: Super :)

lennerd: I will add a comment, to clarify this for future readers. ;)

lennerd:
Öhm … This is not finished nor ready to be merged.

PR #161, opened 2018-10-19
TTL Caches

lennerd:
* Add `TTLCache` class
* Add ClientSideConfigurationCache with TTL
* Add CategoryCache with TTL
* Add TTL to LicenseCache and SourceCache


lennerd:
Do you have more information about the routing? I have a little guess because I tried to fix some errors in the `_app` page. I can imagine the problem is somewhere in this file.

PR #166, opened 2018-11-05
Shrink size of twitter card rendering

mutaphysis:
- do not serialize app properties when rendering for twitter 
- do not render application when rendering for twitter

Intial: 3.3MB
Without Serialized Props: 62KB
Without Page: 12KB

### DIFF
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -33,6 +33,10 @@ import env from '../lib/env';
 import isCordova from '../lib/isCordova';
 
 let isServer = false;
+// only used in serverSideRendering when getting the initial props
+// used for storing the initial props instead of serializing them for the client
+// to prevent sending too large html chunks, that break e.g. twitter cards
+let nonSerializedProps: ?{ appProps: AppProps, routeProps: any | void } = null;
 
 export default class App extends BaseApp {
   static async getInitialProps({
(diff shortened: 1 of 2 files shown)


src/pages/_app.js
```diff
@@ -33,6 +33,10 @@ import env from '../lib/env';
 import isCordova from '../lib/isCordova';
 
 let isServer = false;
+// only used in serverSideRendering when getting the initial props
+// used for storing the initial props instead of serializing them for the client
+// to prevent sending too large html chunks, that break e.g. twitter cards
+let nonSerializedProps: ?{ appProps: AppProps, routeProps: any | void } = null;
```
opyh: 0: If it's hardcoded anyway: How about `propsOmittedForTwitterBot`?

src/pages/_app.js
```diff
@@ -124,9 +128,21 @@ export default class App extends BaseApp {
       return { statusCode };
     }
 
+    // when requested by server side rendering only, skip serializing app props as these are huge
+    const userAgent = appProps.userAgent.ua || '';
+    const isTwitterBot = userAgent.includes('Twitterbot');
+    const doNotSerializeAppProps = isTwitterBot && isServer;
+
+    if (doNotSerializeAppProps) {
```
opyh: 0: How about `shouldSerializeAppProps`?

src/pages/_app.js
mutaphysis: I assume we could also just this for others only interested in the head, so I went for a more generic name.

PR #167, opened 2018-11-06
Allow to set disableWheelmapSource

mutaphysis:
- Hide category menu when disableWheelmapSource=true
- Do not fetch wheelmap data when disableWheelmapSource=true
- Hide wheelmap layer from map when disableWheelmapSource=true

Use with e.g.

http://localhost:3000/?disableWheelmapSource=true&includeSourceIds=QxyZ7vAMS695Bhvii


mutaphysis:
screenshot 2018-11-06 at 11 03 33

PR #172, opened 2018-11-08
Allow overriding lat & lon for the first load

mutaphysis:


src/App.js
```diff
@@ -218,12 +229,7 @@ class App extends React.Component<Props, State> {
       'map.lastMoveDate': new Date().toString(),
     });
 
-    this.setState({
-      extent: null,
-      lat: null,
-      lon: null,
-      zoom: null,
-    });
+    this.setState({ extent: null, lat: lat, lon: lon, zoom: zoom });
```
dakeyras7: You could use property shorthand notation here:

```js
this.setState({ extent: null, lat, lon, zoom });
```

mutaphysis: Good point!

mutaphysis: Fixed

PR #174, opened 2018-11-12
Allow overriding appId from url

mutaphysis:


src/components/MainMenu/MainMenu.js
```diff
@@ -209,6 +209,7 @@ const StyledMainMenu = styled(MainMenu)`
 
   .logo {
     margin-left: 10px;
+    object-fit: contain;
```
dakeyras7: -2: This css rule seems fairly new. It's not supported for IE and older iOS Versions: https://caniuse.com/#search=object-fit

src/components/Onboarding/Onboarding.js
```diff
@@ -208,6 +208,7 @@ const StyledOnboarding = styled(Onboarding)`
     .logo {
       width: 250px;
       height: 53px; // IE 11 does not preserve aspect ratio correctly and needs a fixed height.
+      object-fit: contain;
```
dakeyras7: -2: Same as the comment earlier

src/components/MainMenu/MainMenu.js
mutaphysis: It's a quick fix against scaling logos to a wrong aspect ratio - I guess we need to rethink the custom logo generation we do now. But unrelated to this PR - should I remove the entire line, or is it okay that the look will differ on older browsers?

mutaphysis: Can I use says 90% coverage, so I would be fine with that :D

dakeyras7: I see. As long as it's "degrading gracefully", which in this case would be that the logo image in certain cases is distorted, it should be fine. Also it's a very edgy case tbh :)

mutaphysis: @opyh any opinion? Can you also take a look?

opyh: Damn! No support for IE11 would be a typical issue with bigger companies like banks. We once said we want at least IE11. How about having a div with background-size: cover and a sensible aria label to make up for the bad HTML semantics?

PR #177, opened 2018-11-14
Wheelmap Embed Mode

dakeyras7:
Disclaimer: I couldn't find a more elegant way to deal with the differences of embedded vs non-embedded mode other than to just use conditionals throughout. Code is a little bit distributed, so I wasn't able to abstract it better.

This enables a wheelmap configuration with a query parameter `embedded=true`. When this is set Wheelmap changes in the following ways:

* The main menu is not displayed
* The search toolbar is not displayed
* The node toolbar is uses more space towards the top (since missing main menu and search toolbar opened up more space
* There is a Wheelmap home link that opens the full wheelmap site in a new tab now. The button is positioned on the right on wide viewports and on top for small viewports
* The previous redirect for the classic embed urls now instead redirects to the new wheelmap embed url

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -170,9 +170,13 @@ class App extends React.Component<Props, State> {
   }
 
   componentDidMount() {
+    const { routeName, inEmbedMode } = this.props;
+
+    const shouldStartInSearch = routeName === 'map' && !inEmbedMode;
+
     if (isFirstStart()) {
       this.setState({ isOnboardingVisible: true });
-    } else if (this.props.routeName === 'map' && !this.props.inEmbedMode) {
+    } else if (shouldStartInSearch) {
       this.openSearch(true);
     }
   }
--- a/src/App.js
+++ b/src/App.js
@@ -64,6 +64,7 @@ type Props = {
   lon: ?string,
   zoom: ?string,
   extent: ?[number, number, number, number],
+  isEmbeddedWidget: boolean,
 
   includeSourceIds: Array<string>,
   excludeSourceIds: Array<string>,
@@ -640,6 +641,7 @@ class App extends React.Component<Props, State> {
       isOnSmallViewport: this.state.isOnSmallViewport,
       isSearchToolbarExpanded: this.state.isSearchToolbarExpanded,
       searchResults: this.props.searchResults,
+      isEmbeddedWidget: this.props.isEmbeddedWidget,
 
       disableWheelmapSource: this.props.disableWheelmapSource,
       includeSourceIds: this.props.includeSourceIds,
--- a/src/App.js
+++ b/src/App.js
@@ -172,7 +172,7 @@ class App extends React.Component<Props, State> {
   componentDidMount() {
     if (isFirstStart()) {
       this.setState({ isOnboardingVisible: true });
-    } else if (this.props.routeName === 'map') {
+    } else if (this.props.routeName === 'map' && !this.props.inEmbedMode) {
       this.openSearch(true);
     }
   }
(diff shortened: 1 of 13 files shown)


dakeyras7:
Ready for Review, but no merging please. Needs to be tested in external page first.

src/App.js
```diff
@@ -171,7 +172,7 @@ class App extends React.Component<Props, State> {
   componentDidMount() {
     if (isFirstStart()) {
       this.setState({ isOnboardingVisible: true });
-    } else if (this.props.routeName === 'map') {
+    } else if (this.props.routeName === 'map' && !this.props.inEmbedMode) {
```
mutaphysis: 0: combine into boolean startInSearch?

src/components/WheelmapHomeLink.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { t } from 'ttag';
+import styled from 'styled-components';
+
+const WheelmapHomeLink = ({ className, logoURL }) => (
+  <a
+    className={className}
+    href="https://wheelmap.org"
```
mutaphysis: -1: As you are using the clientSideConfiguration logo, shouldn't this also use the client side configuration hostName and label?

src/components/WheelmapHomeLink.js
```diff
@@ -0,0 +1,24 @@
...
+    aria-label={t`Go to wheelmap.org`}
```
mutaphysis: 0: New translation? Add translator comment.

src/server.js
```diff
@@ -36,14 +36,15 @@ app.prepare().then(() => {
...
       /(\/[a-zA-Z_-]+)?\/community_support\/new.*/,
-      /(\/[a-zA-Z_-]+)?\/embed.*/,
       /(\/[a-zA-Z_-]+)?\/api\/docs.*/,
     ],
     (req, res) => {
       res.redirect(`http://classic.wheelmap.org${req.originalUrl}`);
     }
   );
 
+  server.get([/(\/[a-zA-Z_-]+)?\/embed.*/], (req, res) => res.redirect('/?embedded=true'));
```
mutaphysis: -2: make sure to keep old url params - lat, lon etc are stripped this way.

mutaphysis:
The styling and logic looks straightforward enough!

src/App.js
dakeyras7: Fixed

src/components/WheelmapHomeLink.js
dakeyras7: This is unclear for me though. Is it the aim of the embed mode to bring "wheelmap" to external pages, or should any white-label app be embeddable?

dakeyras7:
You mean the redirect of the embed page is ok, except for the query parameter being thrown away? But other than that we can do the redirect of old embed links to the new embed mode right?

src/server.js
dakeyras7: Fixed

mutaphysis:
> @mutaphysis You mean the redirect of the embed page is ok, except for the query parameter being thrown away? But other than that we can do the redirect of old embed links to the new embed mode right?

Yes, thanks!

src/components/WheelmapHomeLink.js
dakeyras7: I totally agree. I'm just not sure where to get the "home url" from. It seems like we don't provide that in the `clientSideConfiguration` yet. Let me check

dakeyras7: Fixed

src/components/WheelmapHomeLink.js
dakeyras7: Fixed

PR #183, opened 2018-11-26
Fix Place Details Photos Not Appearing on First Render (SSR)

dakeyras7:
Seemed like an easy fix, after a big of digging, but the whole logic around promise resolution with SSR was hard to grasp for me, so better to thoroughly check if I made a broken assumption here.


opyh:
Looks fine. We just have to check if photos are loading fast enough.

PR #184, opened 2018-11-26
Open External Content in New Tab (based on feat/add-create-route)

dakeyras7:
This branch is based on the `feat/add-create-route` branch, because there were some changes to the way the main menu was rendered in the latter.

This makes all links to external pages open in new tabs so that the user doesn't not lose the app state and everything has to reload when they hit the back button.

### DIFF
--- a/src/components/Link/Link.js
+++ b/src/components/Link/Link.js
@@ -66,7 +66,7 @@ class Link extends React.Component<Props> {
   };
 
   renderExternalLink = (to: string) => (
-    <a {...this.props} href={to}>
+    <a {...this.props} href={to} target="_blank" rel="noopener noreferrer">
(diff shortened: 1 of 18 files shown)


src/components/Link/Link.js
```diff
@@ -36,26 +39,50 @@ class Link extends Component<Props> {
...
+    <a {...this.props} href={to} target="_blank" rel="noopener noreferrer">
+      {this.props.children}
+    </a>
+  );
+
   render() {
-    return <RouteConsumer>{this.renderLink}</RouteConsumer>;
+    const { to } = this.props;
+
+    if (isUrl(to)) {
```
opyh: Yep! This should make the thing easier to use.

PR #185, opened 2018-11-26
Reduce Flow Type Errors

dakeyras7:
I collected these small fixes along the way during the last sprint.

Flow errors down from 273 to 251.

Mühsam ernährt sich das Eichhörnchen... 🐿


opyh:
🌰🌰🌰

PR #194, opened 2019-02-01
Place clusters show up in panel

mutaphysis:


src/components/Map/ClusterIcon.js
```diff
@@ -4,56 +4,93 @@ import L from 'leaflet';
...
+import { interpolateWheelchairAccessibility } from '../../lib/colors';
 import isSamePlace from './isSamePlace';
 import getIconNameForProperties from './getIconNameForProperties';
-import * as categoryIcons from '../icons/categories';
+import * as markers from '../icons/markers';
+import Icon, { StyledIconContainer } from '../Icon';
 
 export default class ClusterIcon extends L.Icon {
   constructor(options: typeof L.Icon.options) {
+    // increased tap region for icons, rendered size might differ
```
opyh: -1: Have you tested that this does not accidentally cover PoIs below clusters and makes them unreachable via tap?

src/components/Map/ClusterIcon.js
```diff
@@ -4,56 +4,93 @@ import L from 'leaflet';
...
+      popupAnchor: new L.Point(size * 0.5, size * 0.5),
+      tooltipAnchor: new L.Point(size * 0.5, size * 0.5 + 25),
     };
 
     super(Object.assign(defaults, options));
   }
 
   createIcon() {
-    const div = document.createElement('div');
+    const elem = document.createElement('a');
```
opyh: -1: A11y tests dislike `<a>` without href – how about `<button>`?

src/components/Map/Map.js
```diff
@@ -168,15 +173,30 @@ export default class Map extends React.Component<Props, State> {
...
 
-    const zoom = overrideZoom || fallbackZoom;
-
-    const placeOrEquipment = state.placeOrEquipment;
-
+    let zoom = overrideZoom || fallbackZoom;
     let center = [0, 0];
     let bounds = null;
     let zoomedToFeatureId = state.zoomedToFeatureId;
+    let centerDefined = false;
```
opyh: I have to admit I don't fully understand the logic behind this flag… could you add a comment what it does? Maybe renaming is even enough :)

src/components/NodeToolbar/NodeToolbar.js
```diff
@@ -357,7 +357,7 @@ class NodeToolbar extends React.Component<Props, State> {
         <ErrorBoundary>
           <FocusTrap
             component="div"
-            // We need to set clickOutsideDeactivates here as we want clicks on e.g. the map markers to not be pervented.
+            // We need to set clickOutsideDeactivates here as we want clicks on e.g. the map markers to not be prevented.
```
opyh: …perverted!

src/lib/colors.js
```diff
@@ -85,39 +85,134 @@ const interpolateYesLimited = interpolateLab(
...
-export function interpolateWheelchairAccessibilityColors(propertiesArray: NodeProperties[]) {
+function calculateWheelchairAccessibility(propertiesArray: NodeProperties[]) {
   if (!propertiesArray || propertiesArray.length === 0) {
-    return colors.markers.background.unknown;
+    return { definedCount: 0 };
   }
-  const accessibilityValues = propertiesArray.map(isWheelchairAccessible);
+
+  // take a 'random' selection when there are too many places
+  const filterMod = Math.floor(propertiesArray.length / 30);
```
opyh: -1: lodash has a nicer-looking `sample` method :)

I think it's a good idea, but out of curiosity, did this noticeably change speed?

src/components/Map/Map.css
```diff
@@ -229,75 +229,51 @@ body.is-touch-device .leaflet-control-zoom {
   fill: white;
 }
 
-.ac-marker-current {
```
opyh: +35: Finally!

src/components/Map/ClusterIcon.js
mutaphysis: Just aligned with Icon.js

src/components/Map/ClusterIcon.js
mutaphysis: The same effect was in place before, just in Map.css. Clustering radius takes the increased click area into account, and we never had complaints so far.

src/components/Map/Map.js
mutaphysis: Will do!

src/lib/colors.js
mutaphysis: I didn’t measure really. But I was running with the react inspector plugin, this is pretty slow already. I can add some data on this.

PR #195, opened 2019-02-01
Show cluster in place details

mutaphysis:
Based on #194


src/components/NodeToolbar/AccessibilitySection/PlaceAccessibilitySection.js
```diff
@@ -47,7 +48,7 @@ export default function PlaceAccessibilitySection(props: Props) {
   const descriptionElement = description ? <Description>{description}</Description> : null;
 
   return (
-    <StyledFrame>
+    <StyledFrame noseOffsetX={cluster ? 67 : undefined}>
```
opyh:
```suggestion
    <StyledFrame arrowOffsetX={cluster ? 67 : undefined}>
```

🐽

src/components/NodeToolbar/AccessibilitySection/StyledFrame.js
```diff
@@ -21,7 +21,7 @@ const StyledFrame = styled.div`
     position: absolute;
     content: ' ';
     top: -8px;
-    left: 10px;
+    left: ${props => props.noseOffsetX || 10}px;
```
opyh:
```suggestion
    left: ${props => props.arrowOffsetX || 10}px;
```

opyh:
Problems I found so far when trying this out:

When clicking on a cluster, I get an exception
screenshot 2019-02-01 at 19 10 01

Red clusters should be a horizontal/vertical square, not a diamond:
screenshot 2019-02-01 at 19 08 46

Bigger clusters should be rendered bigger than smaller clusters, as before:
screenshot 2019-02-01 at 19 08 37

Text color should be white:
screenshot 2019-02-01 at 19 08 31

mutaphysis:
I will check the stack overflow, but I have a feeling that’s also the same problem. It happened during hot reloads for me as well.

opyh:
Fixed the crash.

```
if (this.state.zoomedToFeatureId !== targetMapState.zoomedToFeatureId) {
   this.setState({
     zoomedToFeatureId: targetMapState.zoomedToFeatureId,
   });
}
```

sets a state within a `componentDidUpdate` callback that happens after `setState`, leading to an infinite loop.

I've commented this out for now. It's okay for now if the UI zooms again when clicking on a cluster twice IMO.

PR #197, opened 2019-04-11
Added tracking-events

mutaphysis:
- Added new backend for sending tracking events to the accessibility apps 
  backend
- Events are only sent, when cookies/tracking accepted
- A userUUID is generated on firsr usage and removed when tracking rights 
  are revoked
- User position is collected in background when location tracking is active
  on the map

- Missing:
  - Joined event
  - App
  - Organization


src/lib/TrackingEventBackend.js
```diff
@@ -0,0 +1,72 @@
...
+          {
+            method: 'POST',
+            headers: {
+              Accept: 'application/json',
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({
+              ...event,
+              userUUID,
+              timestamp: (Date.now() / 1000) | 0,
```
dakeyras7: -1: How about using `parseInt` or `Math.floor` to get an Integer out of this. Seems more clear to my eyes. I honestly had to look up why you are using the bitwise operator with 0 here 😅

src/lib/TrackingEventBackend.js
```diff
@@ -0,0 +1,72 @@
...
+              userAgent: getUserAgent(),
+              userLocation: userLocation
+                ? [userLocation.coords.longitude, userLocation.coords.latitude]
+                : undefined,
```
dakeyras7: -1: Would it also make sense to assign `null` here instead of `undefined`? I would always prefer `null` if possible

src/lib/UserPositionTracker.js
```diff
@@ -0,0 +1,54 @@
+import { shouldLocate, addListener } from './savedState';
+
+/**
+ * The UserPositionTracker allows tracking changes to the user position.
+ * It respects the global setting for tracking on the map, and only
+ * collects positions while it is turned on.
+ */
+export default class UserPositionTracker {
+  userLocation: null | Position = null;
```
dakeyras7: +0: We could use the optional type here, righ? `userLocation: ?Position = null;`

src/lib/UserPositionTracker.js
```diff
@@ -0,0 +1,54 @@
...
+  handleStorageChange = () => {
+    if (shouldLocate()) {
+      this.start();
+    } else {
+      this.stop();
+    }
+  };
+
+  start() {
+    if (this._watch) {
```
dakeyras7: +0: Curiosity: Why the underscored variable convention?

src/lib/UserPositionTracker.js
```diff
@@ -0,0 +1,54 @@
...
+  stop() {
+    if (this._watch === null) {
+      return;
+    }
+    navigator.geolocation.clearWatch(this._watch);
+    this._watch = null;
+    this.userLocation = null;
+  }
+
+  _watch: number | null = null;
```
dakeyras7: +0: This could also be an optional type, correct?

src/AppContext.js
```diff
@@ -1,16 +1,23 @@
...
+  categories: CategoryLookupTables,
 };
 
-// FIXME
-// The next line should be rather:
-// const { Provider, Consumer } = createContext<AppContext>();
-//
-// Adding the correct generic typing here produces errors during compilation though.
-const { Provider, Consumer } = createContext();
+const { Provider, Consumer } = createContext<AppContext>({
```
dakeyras7: +2: Wonderful. Thx for fixing this. I tried this before and it didn't work. Now my suspect is that I forgot the `// @flow` line at the top.

src/pages/_app.js
```diff
@@ -296,7 +296,11 @@ export default class App extends BaseApp {
 
     const availableLocales: Locale[] = Object.keys(allTranslations).map(localeFromString);
 
-    const appContext = { baseUrl };
+    const appContext = {
+      baseUrl,
+      categories: appProps.categories,
+      // TODO add joined mapping event here
```
dakeyras7: +0: In my current implementation of joined events I'm passing the joined event down the tree. I'm not convinced yet if we should add it to the global `AppContext`. #globalsphobia

src/lib/TrackingEventBackend.js
```diff
@@ -0,0 +1,72 @@
...
+          `${
+            env.public.accessibilityCloud.baseUrl.accessibilityApps
+          }/tracking-events/report?appToken=${env.public.accessibilityCloud.appToken}`,
+          {
+            method: 'POST',
+            headers: {
+              Accept: 'application/json',
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({
```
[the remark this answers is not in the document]
mutaphysis: Will do.

src/lib/TrackingEventBackend.js
mutaphysis: Sure, will change :)

src/lib/TrackingEventBackend.js
mutaphysis: `undefined` will omit the field in the json, `null`would include it. Would rather keep it `undefined`.

src/lib/UserPositionTracker.js
mutaphysis: Not really needed, will change. Was meant to be private :)

src/lib/UserPositionTracker.js
mutaphysis: Yes, would work - optional allows `undefined` & `null`, I did it like this to indicate the null possibility clearer.

src/AppContext.js
mutaphysis: Yes, thats it.

src/pages/_app.js
mutaphysis: Otherwise I will have to pass it through
NodeToolbarFeatureLoader -> NodeToolbar ->  WheelchairStatusEditor -> RadioStatusEditor just for the purpose of sending a tracking event. 

This feels a bit like polluting the hierarchy too much. I could also move the tracking logic into eg. the app, so that I would have all the data available there, no need for context then.

Any opinions? Also @opyh potentially?

src/lib/UserPositionTracker.js
mutaphysis: Commented.

src/lib/UserPositionTracker.js
mutaphysis: Done.

src/lib/UserPositionTracker.js
mutaphysis: Done.

src/lib/TrackingEventBackend.js
mutaphysis: Done.

src/lib/TrackingEventBackend.js
mutaphysis: Done

src/lib/TrackingEventBackend.js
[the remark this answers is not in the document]
mutaphysis: Done

src/pages/_app.js
opyh: I think it's okay to have this there globally as it's one of the few things that are actually a global context affecting everything in the app - on the same level as a potential user auth session.

dakeyras7: Sure. I agree so far. Only thing we need to consider is that we only get the joined mapping event id pretty late (only when we are client-side and therefore local storage is available). So I guess `_app.js` might not be the right place to set it.

opyh: Good point, will consider this when going on with this.

PR #202, opened 2019-05-06
Track `appId` and `organizationId` on a11y editing

opyh:
- Extends the app context by an `app` attribute that contains the whole app’s configuration JSON
- Don’t cache `clientSideConfiguration` anymore, but its parent `App` object
- Make `app` property available over `AppContextConsumer`


src/components/NodeToolbar/AccessibilityEditor/saveStatus.js
```diff
@@ -71,6 +72,7 @@ function save<T>(options: SaveOptions<T>): Promise<Response> {
       throw response;
     })
     .then(json => {
+      debugger;
```
dakeyras7: @opyh This debugger statement is now in master right?

opyh: Oops. Will remove it.

PR #206, opened 2019-05-10
Different Panning/Zooming Input Methods in Embed Mode

dakeyras7:
I found a Plugin for Leaflet which mimics the Google Maps behavior regarding panning and zooming for special cases like touch devices or for embedded map widgets.

A few notes:

* This is directly tied to the `embedded` query parameter. When it is true the mode switches.
* This is not only working on touch but is also useful for embedded widgets when visited on desktop. The text reflects this accordingly. You can see the examples on the [plugin website](https://www.npmjs.com/package/leaflet-gesture-handling): 
* The translations for the overlay text are already packaged with the plugin but there was exactly one instance where the translation for the macOS case was broken, and this was german. So I went ahead and [forked the repo](https://github.com/sozialhelden/Leaflet.GestureHandling) to get to a quick solution.
* While at it (=forking), I also made it so that the text appears center-aligned cause it looks nicer.

I reckon that while this is a good enough solution for now, the project does not seem to be well supported and we might run into other bugs in the future and would have to maintain our fork.
But maybe all this does not matter if we ditch Leaflet in the future anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "is-url": "^1.2.4",
     "isomorphic-unfetch": "^2.1.1",
     "leaflet": "^1.0.3",
+    "leaflet-gesture-handling": "sozialhelden/Leaflet.GestureHandling",
     "leaflet-tilelayer-geojson": "^1.0.4",
     "leaflet.markercluster": "^1.0.4",
     "local-storage-fallback": "^4.1.1",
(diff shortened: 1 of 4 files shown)


dakeyras7:
@opyh @mutaphysis I do have one last doubt/question about this: The plugin comes with its own text translations (52 languages) that are not under our control. Is that ok for us?

mutaphysis:
I think it would be best if we would pass in our own translated strings so that transifex can be used, but this will of course take time & so on.

opyh:
Cool! Totally okay IMO. I'd suggest if we need more languages, we directly make a PR if necessary.

PR #209, opened 2019-05-14
Tracking Events for Opening App and Joining/Leaving Events

dakeyras7:
Three new Event Types exist now:

* `AppOpened` is sent when the app is opened the website is reloaded
* `MappingEventJoined` is sent when the user joins the mapping event either by the join URL or by manually clicking the button. The two cases are distinguishable by the property `joinedVia = 'url' | 'button'` within the event.
* `MappingEventLeft` is sent when the user leaves the event by clicking the leave button on the event detail page. The event comes with an extra property of `leftMappingEventId` that signifies the just left mapping event

Also small change: The `trackingEventBackend.track()` function now expects the app and not the app context.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import includes from 'lodash/includes';
 import findIndex from 'lodash/findIndex';
 import type { Router } from 'next/router';
+import queryString from 'query-string';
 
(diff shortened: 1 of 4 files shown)


opyh:
Could you fix the conflict and press the rebase button? :)

PR #214, opened 2019-05-16
Fix Node Header Issues

dakeyras7:
This PR fixes the upper extra margin on Safari due to different results of using `position: sticky`, by putting the close button within the header and only let the whole header be sticky.

This also solves the "visual gap" between header and button when the toolbar is scrollable.

As another optimization I made the grab-handle narrower in width so we can avoid click target conflicts with the header icon button or the close button.


dakeyras7:
I discovered some new minor glitches that I have to fix, before we merge this.

PR #217, opened 2019-06-03
Event Days Left on Mapping Event Detail

dakeyras7:
What the title says. This was missing before.

I decided for the following presentation:

* If the event ended already we don't shoe any counter
* If the event would end on the current day it reads "1 day left"
* If the event ends tomorrow it reads "2 days left"
* and so on...

I also added another script for ui element svg component generation


src/components/MappingEvents/MappingEventToolbar.js
```diff
@@ -132,20 +140,29 @@ const MappingEventToolbar = ({
...
+          <div className="statistic">
             <div className="statistics-count">
-              <BellIcon />
+              <UserIcon />
               <span>{mappingEvent.statistics.invitedParticipantCount}</span>
             </div>
             <div className="statistics-description">{inviteesCountAriaLabel}</div>
           </div>
+          {remainingDaysLeft && (
+            <div className="statistic">
```
mutaphysis: 0: Wrap in own styled component? It's just a line of css now, so might not be worth it :)

mutaphysis: But it could include all the other statistics-rules.

mutaphysis: Also NITPICK: `statistics` vs `statistic`?

dakeyras7: Yeah, let me see if a separate component is sensible here.

`statistics` means the container and `statistic` each element within. What would you nit-correct? :)

dakeyras7: I had some incosistencies, just realized. The container is now in plural form and the class names for each individitual "statistic" are in singular form now.
Maybe that's what you meant? :)

dakeyras7: I also put the statistics stuff into its own subcomponent. Is a bit more clean this way.

mutaphysis:
Cool, thanks!

PR #218, opened 2019-06-03
Fix Issues Surrounding "Planned" Events

dakeyras7:
There were a couple of things we forgot to add to make "planned" events treated as "ongoing" events.

* Show planned events on the map as well
* Add planned events to the "is event ongoing" logic when checking if local storage value for active event has to be reset

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -270,7 +270,10 @@ class App extends React.Component<Props, State> {
   isMappingEventOngoing(mappingEventId: ?string, mappingEvents: MappingEvents) {
     if (mappingEventId) {
       const joinedMappingEvent = mappingEvents.find(event => event._id === mappingEventId);
-      return joinedMappingEvent && joinedMappingEvent.status === 'ongoing';
+      return (
+        joinedMappingEvent &&
+        (joinedMappingEvent.status === 'ongoing' || joinedMappingEvent.status === 'planned')
+      );
     }
 
     return false;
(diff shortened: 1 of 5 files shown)


src/components/Map/Map.js
```diff
@@ -572,7 +572,12 @@ export default class Map extends React.Component<Props, State> {
 
     this.props.mappingEvents &&
       this.props.mappingEvents
-        .filter(event => event.status === 'ongoing' || event._id === this.props.featureId)
+        .filter(
+          event =>
+            event.status === 'ongoing' ||
+            event.status === 'planned' ||
+            event._id === this.props.featureId
```
mutaphysis: Unrelated: featureId is no longer only a featureId, if I read this correctly?

src/App.js
```diff
@@ -270,7 +270,10 @@ class App extends React.Component<Props, State> {
   isMappingEventOngoing(mappingEventId: ?string, mappingEvents: MappingEvents) {
     if (mappingEventId) {
       const joinedMappingEvent = mappingEvents.find(event => event._id === mappingEventId);
-      return joinedMappingEvent && joinedMappingEvent.status === 'ongoing';
+      return (
+        joinedMappingEvent &&
+        (joinedMappingEvent.status === 'ongoing' || joinedMappingEvent.status === 'planned')
```
mutaphysis: 0: Maybe move into some small helper `isEventVisible() : bool`, that you use in both places

src/components/Map/Map.js
dakeyras7: Good catch: This might be a leftover inconsistency since in the beginning a `MappingEvent` was in indeed a `Feature`. It isn't anymore. There is now a `MappingEventFeature` type. I will probably clean this up in the current trello ticket were I clean up event-related typings

PR #219, opened 2019-06-04
Fix Days Left Calculation

dakeyras7:
Just realized I was doing it wrong 🤦🏻‍♂️

`endDate - currentDate` sound way more plausible then `endDate - startDate`

Could you please check that I'm not wrong again plz?

### DIFF
--- a/src/components/MappingEvents/Statistics.js
+++ b/src/components/MappingEvents/Statistics.js
@@ -33,7 +33,7 @@ const Statistics = ({
   const currentDate = new Date();
   const remainingDaysLeft =
     endDate && endDate > currentDate
-      ? Math.floor((endDate - currentDate) / 1000 / 60 / 60 / 24) + 1
+      ? Math.ceil((endDate - currentDate) / 1000 / 60 / 60 / 24)
       : null;
 
   return (
--- a/src/components/MappingEvents/Statistics.js
+++ b/src/components/MappingEvents/Statistics.js
@@ -30,9 +30,10 @@ const Statistics = ({
   // translator: Description for number of days left in the mapping event
   const daysLeftLabel = t`Verbleibende Tage`;
 
+  const currentDate = new Date();
   const remainingDaysLeft =
-    endDate && endDate > new Date()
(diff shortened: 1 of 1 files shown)


src/components/MappingEvents/Statistics.js
```diff
@@ -30,9 +30,10 @@ const Statistics = ({
   // translator: Description for number of days left in the mapping event
   const daysLeftLabel = t`Verbleibende Tage`;
```
mutaphysis: Unrelated: did we switch to german at some point?

src/components/MappingEvents/Statistics.js
```diff
@@ -30,9 +30,10 @@ const Statistics = ({
...
 
+  const currentDate = new Date();
   const remainingDaysLeft =
-    endDate && endDate > new Date()
-      ? Math.floor((endDate - startDate) / 1000 / 60 / 60 / 24) + 1
+    endDate && endDate > currentDate
+      ? Math.floor((endDate - currentDate) / 1000 / 60 / 60 / 24) + 1
```
mutaphysis: Isn't `Math.ceil((endDate - currentDate) / 1000 / 60 / 60 / 24)` what you want, so `ceil` instead of the +1

src/components/MappingEvents/Statistics.js
dakeyras7: No. Probably my oversight. Do we always go english by default?

src/components/MappingEvents/Statistics.js
dakeyras7: Yes! That is true. Will change.

src/components/MappingEvents/Statistics.js
dakeyras7: If we change this now, until this will get translated in transifex, it will show english text though. It that acceptable for a period of time? Our mapping event users currently are only in Germany?

dakeyras7: I will change this to english by default

PR #222, opened 2019-06-11
Make Map Not Locate On Start When Lat Lon Is Specified

dakeyras7:
This fixes: [https://trello.com/c/GukAlO7p/93-wheelmap-widget-mit-fester-lat-lon-startbar](https://trello.com/c/GukAlO7p/93-wheelmap-widget-mit-fester-lat-lon-startbar)

The fix is not specific too widget mode but applies for the app in general. When the lat & lon are specified in the URL the map should never jump away from that the position on load.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -889,11 +889,10 @@ class App extends React.Component<Props, State> {
     const { isSpecificLatLonProvided } = this.state;
     const isNodeRoute = Boolean(this.props.featureId);
     const isNodeToolbarDisplayed = this.isNodeToolbarDisplayed();
+    const mapMoveDate = savedState.map.lastMoveDate;
+    const wasMapMovedRecently = mapMoveDate && new Date() - mapMoveDate < config.locateTimeout;
 
-    const shouldLocateOnStart =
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -882,11 +886,14 @@ class App extends React.Component<Props, State> {
...
   render() {
+    const { isSpecificLatLonProvided } = this.state;
     const isNodeRoute = Boolean(this.props.featureId);
     const isNodeToolbarDisplayed = this.isNodeToolbarDisplayed();
 
     const shouldLocateOnStart =
-      !isNodeRoute && +new Date() - (savedState.map.lastMoveDate || 0) > config.locateTimeout;
+      !isSpecificLatLonProvided &&
+      !isNodeRoute &&
+      +new Date() - (savedState.map.lastMoveDate || 0) > config.locateTimeout;
```
mutaphysis: Unrelated, can you extract this to another variable as well, kinda hard to read :)

dakeyras7: I cleaned it up now with a `wasMapRecentlyMoved` variable and also got rid of the `+` and optional `0`

PR #227, opened 2019-06-17
Mapping Event Toolbar UI Improvements

dakeyras7:
This fixes a couple of things:

* The meeting point, area and dates are now prefixed with icons to make their purpose clearer
* The dates are now adapted to the optionality of both date fields: When both are displayed they are joined with a `-`.  When only the start date exists we prefix it with `starting` and when only the end date exists we prefix it with `until`. Both prefixes need to be translated. This is not the best way since not all languages work this way, but I didn't want to include more i18n complexity for now.
* The DOM is a bit more clean semantically
* Each of the mentioned elements now have a tootltip and better a11y names


dakeyras7:
Which links? 🤔

opyh:
I meant SVG using global anchors as reference…

dakeyras7:
Yeah those are generated by svgr. I forgot to mention that I had to update the library for SVGR,  because the old version would generate different SVGs that were using the same anchors which caused bugs

PR #233, opened 2019-07-01
Display correct statistics in front-end

mutaphysis:
Field names for statistics changed since the original implementation, this moves the values
to the correct implementation


dakeyras7:
LGTM ;)

PR #238, opened 2019-07-08
Removed all remaining cordova remainders

mutaphysis:


src/components/ShareBar/ShareBar.js
```diff
@@ -101,14 +100,7 @@ class ShareBar extends React.Component<Props, State> {
 
     if (!this.state.isExpanded) return expandButton;
 
-    // The share button lib we're using is using window.open() by default which
-    // seems to not work in our Cordova app for some reason. If we are a Cordova
-    // app then we configure the share buttons to use window.location.href which
-    // works in Cordova.
-    // If we are on web we just use the default behavior of the share buttons.
-    const linkOpeningViaLocationHrefProps = isCordova()
```
opyh: Do the share buttons still work correctly? If not, should we switch them off and replace them with native the sharing APIs?

src/lib/goToLocationSettings.js
```diff
@@ -10,30 +9,18 @@ export default function goToLocationSettings() {
...
 
-  if (isCordova()) {
-    if (
-      userAgent.os.name === 'Android' &&
-      typeof window !== 'undefined' &&
-      window.cordova &&
-      window.cordova.diagnostics &&
-      typeof window.cordova.diagnostics.switchToLocationSettings === 'function'
-    ) {
-      window.cordova.diagnostics.switchToLocationSettings();
```
opyh: Hmm, sure we don't need to keep this as location handling is done natively?

opyh:
Seeing deleted code always makes me happy :) I've created tickets for the remaining 2 questions, this can be merged IMO.

src/lib/goToLocationSettings.js
mutaphysis: This will be done by the native apps, they will hook into the urls used here

src/components/ShareBar/ShareBar.js
mutaphysis: On iOS everything but the e-mail share link work fine, will look into it.

src/lib/goToLocationSettings.js
mutaphysis: iOS is already done, https://github.com/sozialhelden/wheelmap-native-wrapper/commit/876dc653bdfe321983cd0fd7f6d0e6f93575e847

PR #240, opened 2019-10-29
Aligned mapping events filter

mutaphysis:
- Moved filtering into app
    - removed from toolbar/map
   - Always filter by app id & isVisible
   - Removed odd requery-logic from App (is already happening higher up)
   - Display current selected mapping event on map when opened by url
    (e.g. preview mode from accessibility.apps)

Update mapping events on map
   - Map markers where never updated before, now the markers
    are updated whenever the filter or available events
    change

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -133,6 +133,17 @@ function isStickySearchBarSupported() {
   return hasBigViewport() && !isTouchDevice();
 }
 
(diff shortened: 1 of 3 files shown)


mutaphysis:
Oct-29-2019 15-44-00

PR #242, opened 2019-10-31
Fix for IE 11-compatible Babel transpilation

opyh:
This cleans up browserslist / `babel/preset-env` configuration and fixes IE11 by moving `babel-plugin-ttag` out of the way when Next.js / Webpack / Babel transpiles our code.

Here is a more detailed description of the issue: https://github.com/ttag-org/babel-plugin-ttag/issues/138

For fun, it includes little changes to remove deprecation warnings in Next 9.


.babelrc
```diff
@@ -1,15 +1,34 @@
 {
-  "presets": ["next/babel"], // https://goo.gl/aqbWJY
+  "presets": [[
+    "next/babel",
+    {
+      "preset-env": {
+        "useBuiltIns": "entry",
+        "corejs": 3,
```
opyh: Newer than the default. Default would be 2, also this line is necessary to get rid of a warning.

.babelrc
```diff
@@ -1,15 +1,34 @@
...
+        "debug": true
+      },
+      "transform-runtime": {
+        "corejs": 3,
+        "useESModules": false
```
opyh: Without this, the server code would crash on start.

.babelrc
```diff
@@ -1,15 +1,34 @@
...
+  ]], // https://goo.gl/aqbWJY
   "plugins": [
     "transform-flow-strip-types",
-    [
-      "ttag",
-      {
+    "babel-plugin-styled-components"
+  ],
+  "env": {
+    "ttag": {
```
opyh: This is the configuration used when running Babel with `BABEL_ENV=ttag`.

package.json
```diff
@@ -143,9 +147,9 @@
     "version": "echo TODO: Add version syncing to bundle configurations for native apps + git commit -am 'Version bump' to package.json",
     "flow": "flow",
     "server": "server",
-    "push-translations": "npm run clear-babel-cache && npm run build && tx push -s",
+    "push-translations": "npm run clear-babel-cache && (BABEL_ENV=ttag npm run build) && tx push -s",
     "pull-translations": "tx pull -af",
-    "retranslate": "npm run build && npm run pull-translations && NEW_SOURCE_PO='./src/static/i18n/en_US.txt' retranslate/run.sh",
```
opyh: `static` is deprecated

package.json
```diff
@@ -184,7 +188,8 @@
     "Opera >= 53",
     "iOS >= 10",
     "last 1 version",
-    "> 1%"
+    "> 0.25%",
+    "not dead"
```
opyh: Support a few more browsers because why not :)

src/pages/_app.js
```diff
@@ -313,72 +313,70 @@ export default class App extends BaseApp {
     };
 
     return (
-      <Container>
```
opyh: Next.js 9 doesn't need this component wrapper anymore. Everything below this line in this file is just Prettier's re-indentation, feel free to ignore :)

.babelrc
mutaphysis: did you test the pot file extraction with these changes?

opyh: I didn't look in detail at the `.pot` file, but it was generated correctly.

PR #245, opened 2020-03-26
Create place flow v3

mutaphysis:
Yet another create place flow.

https://youtu.be/B7P_VPio_FY

wheelmap-created-places

### DIFF
--- a/src/components/CreatePlaceFlow/components/usePlaceSearch.js
+++ b/src/components/CreatePlaceFlow/components/usePlaceSearch.js
@@ -14,7 +14,7 @@ function usePlaceSearch(minimalRequiredLetterCount: number = 2) {
   React.useEffect(() => {
     if (searchString.length >= minimalRequiredLetterCount) {
       setSearchState('Loading');
-      // TODO pass current geo coords here - do we have a useUserCoords already?
+      // Maybe pass current geo coords here - do we have a useUserCoords already?
       const searchResultPromise = searchPlacesDebounced(searchString, {});
       searchResultPromise
         .then(results => {
--- a/src/components/CreatePlaceFlow/components/usePlaceSearch.js
+++ b/src/components/CreatePlaceFlow/components/usePlaceSearch.js
@@ -14,7 +14,8 @@ function usePlaceSearch(minimalRequiredLetterCount: number = 2) {
   React.useEffect(() => {
     if (searchString.length >= minimalRequiredLetterCount) {
       setSearchState('Loading');
-      // Maybe pass current geo coords here - do we have a useUserCoords already?
+      // intentinally use empty coords
+      // see discussion at https://github.com/sozialhelden/wheelmap-frontend/pull/245#discussion_r398698050
       const searchResultPromise = searchPlacesDebounced(searchString, {});
       searchResultPromise
         .then(results => {
--- a/src/components/CreatePlaceFlow/components/usePlaceSearch.js
+++ b/src/components/CreatePlaceFlow/components/usePlaceSearch.js
@@ -1,0 +1,49 @@
+// @flow
+import * as React from 'react';
+import { searchPlacesDebounced } from '../../../lib/searchPlaces';
+import type { SearchResultCollection } from '../../../lib/searchPlaces';
+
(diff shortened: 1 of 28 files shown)


src/components/CreatePlaceFlow/components/usePlaceSearch.js
```diff
@@ -0,0 +1,49 @@
...
+function usePlaceSearch(minimalRequiredLetterCount: number = 2) {
+  const [searchString, setSearchString] = React.useState<string>('');
+  const [searchState, setSearchState] = React.useState<SearchState>('TypeMore');
+  const [searchResults, setSearchResults] = React.useState<SearchResultCollection | null>(null);
+  const request = React.useRef<Promise<SearchResultCollection> | null>(null);
+
+  React.useEffect(() => {
+    if (searchString.length >= minimalRequiredLetterCount) {
+      setSearchState('Loading');
+      // Maybe pass current geo coords here - do we have a useUserCoords already?
```
opyh: The problem with photon's lat/lon query would be that it gives places somewhere else no priority at all, so you don't find that stuff anymore. We stumbled upon this when implementing general search. We'd have to mix/sort results from 2 queries (one nearby, one by name) to overcome this problem, which I find too complex - so I'd suggest we use the results that work for both the nearby case and the somewhere-else case, which are the ones without geo coordinates. They're not optimal as they show totally irrelevant stuff too (it's possible but highly unlikely that somebody tags something outside their country), but that's better than not finding something that you search at all.

src/components/CreatePlaceFlow/pages/PointGeometryPicker.jsx
```diff
@@ -0,0 +1,164 @@
+// @flow
+import * as React from 'react';
+import styled from 'styled-components';
+import { t } from 'ttag';
+import ReactMapGL, { GeolocateControl } from 'react-map-gl';
```
opyh: +0: What do you think about loading this with `dynamic()` as we don't have Next.js page based code splitting yet?

src/lib/searchPlaces.js
```diff
@@ -35,6 +35,37 @@ export type SearchResultCollection = {
...
+  if (searchResultProperties.osm_type === 'W') {
+    osmId = -osmId;
+  }
+
+  return osmId;
+}
+
+export function buildOriginalOsmId(searchResultProperties?: SearchResultProperties) {
+  if (!searchResultProperties) {
+    return 'osm://unknown/unknown';
```
opyh: -1: Maybe I overlook something, but is there a reason why this is not `undefined`? As the string value's content is not typable, a developer that uses the `originalId` value at another end might overlook this case - if the only cases are '`undefined`' and 'defined + the value actually points to something', this could be safer.

src/pages/_document.js
```diff
@@ -43,24 +44,30 @@ export default class MyDocument extends Document {
                 https://sozialhelden.matomo.cloud
                 https://cdn.matomo.cloud
                 http://cdn.matomo.cloud
+                https://api.mapbox.com
+                https://events.mapbox.com
 
+                http://localhost:4000
```
opyh: This probably belongs in `REACT_APP_ALLOW_ADDITIONAL_DATA_URLS` in the dev env :)

mutaphysis: Oh yes, missed this while removing my local workarounds, will fix

src/lib/searchPlaces.js
mutaphysis: This will not happen in real life, I think our types are a bit lenient. Will change to undefined.

src/components/CreatePlaceFlow/components/usePlaceSearch.js
mutaphysis: Displaying the distance to a search result might also help here.

src/components/CreatePlaceFlow/pages/PointGeometryPicker.jsx
mutaphysis: I'll leave that as an improvement on top, have to restructure the components a bit.

PR #246, opened 2020-03-27
Allow rating of AC Places

mutaphysis:
Modified the existing NodeToolbar and a11y editors to support editing of ac places that allow rating.
Mostly no UI changes, just passing correct values through to the right places and evaluating more than just wheelmap classic objects. Tried to touch as little existing code as possible, but will need some merging with @saryekad changes.

Changed filter logic, so that places without at least an empty accessibility object are now shown on the map.

https://www.youtube.com/watch?v=5a-rzbSZjFE

---

Found a few regressions from previous partial refactorings or just old broken code. Toilets nearby for AC places were not loading (fixed in `c814dac`). Place icons were not correctly updated when the feature changed. Places with unknown categories were not shown on the map at all. 
Added tickets for two more problems https://trello.com/c/MDZKDTQg, https://trello.com/c/92zjsCBV


src/components/NodeToolbar/NodeToolbar.js
```diff
@@ -258,26 +260,38 @@ class NodeToolbar extends React.Component<Props, State> {
   }
 
   renderInlineWheelchairAccessibilityEditor() {
```
opyh: @saryekad seems to have removed this code while you have extended it: https://github.com/sozialhelden/wheelmap-frontend/pull/247#pullrequestreview-384237305 - could you sync this?

src/lib/cache/AccessibilityCloudFeatureCache.js
```diff
@@ -101,6 +101,52 @@ export default class AccessibilityCloudFeatureCache extends FeatureCache<
...
+        )
+        .then((response: Response) => {
+          if (response.ok) {
+            response
+              .json()
+              .then(json => {
+                resolve(json.success);
+              })
+              .catch(reject);
+          } else if (response.json) {
```
opyh: Is there a case where response.json is undefined?

src/lib/getToiletsNearby.js
```diff
@@ -129,7 +138,14 @@ export function fetchToiletsNearFeature(
...
   const [lon, lat] = coords;
 
-  return fetchToiletsNearby(lat, lon, disableWheelmapSource, includeSourceIds, excludeSourceIds);
+  return fetchToiletsNearby(
+    lat,
+    lon,
+    disableWheelmapSource,
+    includeSourceIds,
+    excludeSourceIds,
+    appToken
```
opyh: Yay :D

src/lib/cache/AccessibilityCloudFeatureCache.js
mutaphysis: Yes, if the requests aborts due to cors problems in ff

src/components/NodeToolbar/NodeToolbar.js
mutaphysis: I am still waiting for @saryekad to go into master before rebasing. The whole pr will likely change.

mutaphysis: I was waiting, will rebase today :)

mutaphysis: Done, was easy enough :)

mutaphysis:
rebased against master, @opyh can you take another look

src/app/PlaceDetailsData.js
```diff
@@ -131,9 +131,11 @@ function fetchPhotos(
             return [];
           })
       : Promise.resolve([]),
-  ]).then((photoArrays: PhotoModel[][]) => {
-    return [].concat(photoArrays[0], photoArrays[1]);
-  });
+  ])
+    .then((photoArrays: PhotoModel[][]) => {
+      return [].concat(photoArrays[0], photoArrays[1]);
```
opyh: Eagerly waiting for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat…

src/components/Map/Map.js
```diff
@@ -951,15 +951,9 @@ export default class Map extends React.Component<Props, State> {
 
   createMarkerFromFeature = (feature: Feature, latlng: [number, number]) => {
     const properties = feature && feature.properties;
-    if (!properties) return null;
-    if (
-      !isWheelmapFeature(feature) &&
-      !properties.accessibility &&
-      !includes(EquipmentCategoryStrings, properties.category)
-    ) {
+    if (!properties) {
```
opyh: -1: This makes some unrateable sources visible that should not be displayed… We have some sources that are not rateable, but that have places that are reachable over the UI from elevator details screens. Concept-wise this is not beautiful, but it solved the problem of navigating between elevators in the same station without (1) cluttering the map with duplicate stations and (2) without a need to merge data.

Could this do a check that the source is rateable?

src/components/Map/Map.js
```diff
@@ -863,8 +863,8 @@ export default class Map extends React.Component<Props, State> {
     // them in one array:
     const ids = [!props.equipmentInfoId && props.featureId, props.equipmentInfoId]
       .concat(similarEquipmentIds)
-      .map(String)
-      .filter(Boolean);
+      .filter(Boolean)
+      .map(String);
```
opyh: Ouch

src/components/Map/Map.js
opyh: +1 Revised my opinion, let's put this in a new PR.

mutaphysis: I only care about the `!properties.accessibility` line, can leave in the rest.

mutaphysis: But we don't have sources here, so this would make the map even more horrible.

mutaphysis: I'd rather change the filter to remove only the items we don't want, instead of removing too much or building weird white lists of which elements should be shown, but I'm not sure to which places/sources this applies.

opyh: Trello ticket: https://trello.com/c/SfuLlhUu/161-dont-show-gray-places-from-unrateable-sources

How about showing all places that either have

- existing accessibility
- have no accessibility, but
  -`source.isA11yRatingAllowed`
  - are equipment

?

PR #247, opened 2020-03-29
New Detail Panel

saryekad:
This implements the trello ticket: https://trello.com/c/LS25iM1D/127-rework-detail-page

Only places are shown in this new panel. Equipment is still displayed the old way. For now the panel is full screen but this should be quite easily changeable if we decide to have the app show part panel and part map.


src/components/NodeToolbar/NodeToolbar.js
```diff
@@ -257,30 +254,6 @@ class NodeToolbar extends React.Component<Props, State> {
     );
   }
 
-  renderInlineWheelchairAccessibilityEditor() {
```
opyh: @saryekad This has to be merged with @mutaphysis' new code here: https://github.com/sozialhelden/wheelmap-frontend/pull/246/files#diff-5657e5c46dbcbe59ddafaa650f1f828eR262

PR #250, opened 2020-04-02
 Reload equipment layer once every hour

mutaphysis:
This PR solves a few of the problems with outdated equipment infos (see the detailed analysis in https://trello.com/c/WhgMMtfg/112-fix-grey-elevators-after-leaving-tab-open-caused-by-caching)
being displayed in the app when returning to an old session.

Instead of never touching the map tiles again, the map tiles are
reloaded after half the time it takes to become outdated. As the
information updates every 10minutes on ac, the current 60 minutes
should be fine.

This is cheaper than the opposite approach of reloading all the
elements in the EquipmentInfoCache, as only the visible tiles will
be loaded.

This change does not fix the issue that the place details stay
unrefreshed until a change is triggered on the site. This requires
another timer in the place details, or a change to the feature cache
as so far events are not fired for map tile loads.


src/components/Map/GeoJSONTileLayer.js
```diff
@@ -160,6 +183,7 @@ class GeoJSONTileLayer extends TileLayer {
       x: coords.x,
       y: coords.y,
       z: this._getZoomForUrl(),
+      c: this._cacheBuster,
```
opyh, line 186: You've probably put more thoughts into this so I might have overlooked something… from what I see this shouldn't be necessary - the cache headers for this should be set correctly on the server side so the browser doesn't store tiles for too long, and fastly caches are purged when something changes. Adding this parameter would add responses to the varnish storage that are not reusable by anybody as the tiles have a long TTL there, and reduce the hit rate. Isn't there an alternative for this, e.g. a parameter we can set in the request options so the browser cache is not used for sure? How about setting a cache buster boolean flag that enables this behavior and is reset directly after?

mutaphysis: My browser was aggressively caching equipment map tiles, sadly. Either some of the fastly logic is wrong, or leaflet is doing something weird, or something else is misconfigured. I was still getting the same grey tiles without the cache buster :/

opyh: Thanks for debugging + info! I'll take over from here and try to fix this.

PR #251, opened 2020-04-21
Hide parking a11y description

bjoernuhlig:
Do not show paring a11y description as parking spots may not have doors


src/components/NodeToolbar/AccessibilitySection/WheelchairAndToiletAccessibility.js
```diff
@@ -179,10 +181,11 @@ class WheelchairAndToiletAccessibility extends React.Component<Props> {
     if (!hasContent) {
       return null;
     }
-
+    const showDescription = categoryId !== 'parking';
```
opyh, line 184: 👍

PR #330, opened 2021-11-27
convert NotFound.tsx to fc component

Lulezimukaj:
Notfound.tsx converted from class component to functional component

### DIFF
--- a/src/components/NotFound/NotFound.tsx
+++ b/src/components/NotFound/NotFound.tsx
@@ -1,6 +1,7 @@
 import { t } from 'ttag';
 import get from 'lodash/get';
 import * as React from 'react';
+import { useEffect } from 'react';
 import { findDOMNode } from 'react-dom';
 import styled from 'styled-components';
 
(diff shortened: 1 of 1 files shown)


src/components/NotFound/NotFound.tsx
```diff
@@ -11,106 +12,98 @@ import Logo from '../../lib/Logo';
...
   };
 
-  componentDidMount() {
-    this.focus();
-  }
+  let closeButton: Element | Text | null;
 
-  focus() {
-    if (!this.closeButton || !(this.closeButton instanceof HTMLElement)) {
+  const focus = () => {
```
opyh, line 31: same as above

src/components/NotFound/NotFound.tsx
```diff
@@ -11,106 +12,98 @@ import Logo from '../../lib/Logo';
...
+  statusCode: number;
 };
 
-class NotFound extends React.Component<Props> {
-  closeButton: Element | Text | null;
+const NotFound: React.FC<Props> = ({ className, onReturnHomeClick, statusCode }) => {
+  useEffect(() => focus(), []);
 
-  manageFocus = ({ nativeEvent }) => {
+  const manageFocus = ({ nativeEvent }) => {
```
opyh, line 23: This function be wrapped with `React.useCallback(…)`, otherwise the function object will be re-generated on each rendering.

opyh (approved):
LGTM except for a small enhancement. We will just do the change and merge it. Thanks for your work and time! <3

src/components/NotFound/NotFound.tsx
```diff
@@ -11,106 +12,98 @@ import Logo from '../../lib/Logo';
...
     if (nativeEvent.key === 'Tab') {
       nativeEvent.preventDefault();
     }
   };
 
-  componentDidMount() {
-    this.focus();
-  }
+  let closeButton: Element | Text | null;
```
opyh, line 29: You can use `React.useRef()` here instead of creating the reference function yourself.

PR #331, opened 2021-11-27
Feature/refactor class to fc

janDo-dev:
Converted Onboarding.tsx from class component to functional component.

I had to remove the "NEXT_TELEMETRY_DISABLED=1" command from the dev script in package.json because I got the following error message:

_Der Befehl "NEXT_TELEMETRY_DISABLED" ist entweder falsch geschrieben oder konnte nicht gefunden werden._

For context: The error accured while `npm run dev` on a windows system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,7 +170,7 @@
     "url-loader": "^1.1.2"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


bjoernuhlig (approved):
lgtm, many thanks for refactoring

PR #344, opened 2022-01-27
change html lang-tag string received from preferredLocaleString

bjoernuhlig:
Onboarding.tsx -> `marked` import statement threw type error when running locally, changed to named import
_document.tsx -> `locale` property was always undefined, probably due to recent or older changes. locale is now taken from `preferredLocaleString` property

browser locale is read properly now. not tested for mobile yet


opyh:
LGTM! Let's see if this works on staging.

PR #426, opened 2024-08-27
Update .eslintrc.json

DarkMio:
Minor changes to be less quirky on react uses
- default exports are optional
- FCs can declared however you like it
- react does not have to be imported, if it's using components anyway
- comments can be a bit longer, still
- default properties don't have to be passed, if you don't want them
- allowing spreading in react component properties

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,19 +9,24 @@
     "jsx-a11y"
   ],
   "rules": {
+    "import/prefer-default-export": "off",
     "import/extensions": "off",
+    "react/function-component-definition": "off",
+    "react/react-in-jsx-scope": "off",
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -9,19 +9,24 @@
...
+      "comments": 120,
       "ignoreUrls": true,
       "tabWidth": 2
-    }],
-    "indent": ["error", 2],
-    "quote-props": ["error", "as-needed"],
-    "@stylistic/js/max-len": ["error", {
-      "code": 140,
-      "comments": 80,
+      }],
```
mutaphysis, line 24: Illegaler Indent! ;)

mutaphysis (approved):
lgtm

mutaphysis (approved):

PR #431, opened 2024-08-28
A11ymap improve data loading

opyh:
This should improve data loading and fix loading data inside the cluster.

The PR comes with a few new methods for generalized data loading from AC:

```typescript
const response = useCollectionSWR<IAccessibilityAttribute>({
  collectionName: 'AccessibilityAttributes',
  params: new URLSearchParams({
    limit: '10000',
    appToken,
    surrogateKeys: 'false',
  }),
});
```

```typescript
const source = useDocumentSWR<ISource>({ collectionName: 'Sources', _id: featureId });
```

```typescript
fetchDocumentWithTypeTag<IApp>(baseUrl, 'ac:App', appIdForHostname(hostname), `appToken=${appToken}`);
```

…and adds `EnvContext` for accessing environment variables correctly. This is to remove `process.env` references on the client which Next.js would hard-code into the compiled code - I want to be able to change them without rebuilds for the app to be a bit more [12 factor](https://12factor.net/config).


src/lib/fetchers/ac/useAccessibilityCloudAPI.tsx
```diff
@@ -0,0 +1,8 @@
+import { useEnvContext } from "../../context/EnvContext";
+
+export default function useAccessibilityCloudAPI({ cached = true }: { cached: boolean }) {
+  const env = useEnvContext();
+  const baseUrl = cached ? env.NEXT_PUBLIC_ACCESSIBILITY_CLOUD_BASE_URL : env.NEXT_PUBLIC_ACCESSIBILITY_CLOUD_UNCACHED_BASE_URL;
+  const appToken = env.NEXT_PUBLIC_ACCESSIBILITY_CLOUD_APP_TOKEN;
```
mutaphysis, line 6: should this not come from the app? (not sure anymore, honestly asking)

src/lib/util/useSWRWithErrorToast.tsx
```diff
@@ -34,7 +34,8 @@ export default function useSWRWithErrorToast<Data, ErrorType>(
   transform: ErrorTransformerProps<ErrorType>,
```
mutaphysis, line 34: transform is a pretty confusing name for an error renderer

src/lib/fetchers/ac/useCollectionSWR.tsx
```diff
@@ -0,0 +1,50 @@
...
+  params,
+  cached = true,
+  shouldRun = true,
+}: Args): SWRResponse<ListOrFeatureCollection<D, R, G>, AccessibilityCloudError> {
+  const { baseUrl, appToken } = useAccessibilityCloudAPI({ cached });
+  const paramsWithAppToken = new URLSearchParams(params);
+  paramsWithAppToken.append('appToken', appToken);
+  const paramsString = paramsWithAppToken.toString();
+  const transform = useMemo(() => ({
+    summary: (text: any) => <span>{text}</span>,
```
mutaphysis, line 39: using any is frowned upon, and not every any can be used as a ReactNode (e.g. an actual Error object)

mutaphysis (commented):
I think the idea of having access to the AC APIs like this is nice.

src/lib/fetchers/ac/fetchApp.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export function appIdForHostname(hostname: string) {
+  const cleanedHostName = hostname
+    // Allow sharing a deployment with ngrok
+    .replace(/.*\.ngrok(?:-free)\.app$/, 'wheelmap.tech')
+    // Allow branch test deployments
+    .replace(/.*\.wheelmap\.tech$/, 'wheelmap.tech')
+    // Use 'localhost' as hostname for loopback IP
+    .replace(/127\.0\.0\.1/, 'localhost')
```
DarkMio, line 15: This is a very, very minor thing for an ancient edge case, but some versions of Windows don't resolve `localhost` as `127.0.0.1` (something like Win 7/8 home or so). If the benefits of your setup outweigh it, just keep it like that.

src/lib/fetchers/ac/useCollectionSWR.tsx
opyh: Will remove this whole concept and simplify it.

src/lib/util/useSWRWithErrorToast.tsx
opyh: True!

src/lib/fetchers/ac/useAccessibilityCloudAPI.tsx
opyh: Yes, true!

PR #457, opened 2024-12-03
fix: link to next accessible toilet

kriskbx:
This fixes the link to the next accessible toilet in a11ymap that will be shown if a place doesn't have one. The underlying issue was a wrongly named query parameter.

The `id` returned from the osm-api backend service for the next accessible toilet contains a slash (e.g. `way/1234`), which is currently not supported by the routing configuration which expects `way:1234`. To make it work, I replaced it with a colon as a temporary solution. In the future, I think we should take a look at routing and make sure a route like `/amenities/way/1234` also works.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208533441124221


opyh (approved):
Thanks! 💯

PR #458, opened 2024-12-04
fix: legacy link to next accessible toilet

kriskbx:
This fixes the link to the next accessible toilet in the current version of the Wheelmap. I'm not 100% sure as I'm quite new to the project and codebase, but it seems like the query parameters of the osm-api have changed in the past and were not updated for the nearby toilets API call.


opyh (approved):

opyh:
Yep, that's it - thanks!

PR #459, opened 2024-12-09
Add Radix UI to A11yMap

opyh:
This adds [Radix UI](https://www.radix-ui.com) as component library to A11yMap.

The new code improves dark mode, includes initial Radix UI implementations, simplifies the `<ErrorBoundary />` component, and some other improvements, mostly styling. Some UI parts look less polished with the new code, but IMO we should fix this later to get the other branches on the same codebase early, and to have all those nice new components available.


src/components/shared/useDarkMode.tsx
```diff
@@ -1,62 +1,25 @@
...
-      document.body.classList.add('bp5-light')
-      document.body.classList.remove('bp5-dark')
-    }
-  }, [])
-
-  React.useEffect(() => {
-    if (window.matchMedia === undefined) {
-      return
-    }
+  const { theme, setTheme } = useTheme()
```
kriskbx, line 10: **Non-issue:** Very nice, I like the usage of next/themes 🫶

src/components/App/MainMenu/MainMenu.tsx
```diff
@@ -303,104 +203,113 @@ const StyledNav = styled.nav`
...
+        <div className="claim">{claim}</div>
+
+        <div id="main-menu" role="menu">
+          <AppLinks />
+        </div>
+
+        {closeButton}
+      </StyledNav>
+    </Card>
+	);
```
kriskbx, line 314: **Non-blocker:** Note: I don't want to open a whole new discussion about tabs versus spaces.  I have a very hard time reading this diff and figuring out what has actually changed here. What do you @Mayaryin @opyh think of setting the biome configuration for indentation back from to tabs to spaces for now in order to keep our diffs small with less noise? If we decide to go with tabs we can still do it later in a separate PR and then let biome fix all the files at once. I think it could speed up PR reviews if we postpone this formatting change.

kriskbx (changes requested):
Thanks a ton for your amazing work and PR 🫶 

In my opinion switching to radix is a good idea, I love it. I found a few legacy things that could still be ported to radix while sipping through the code, but I think we can postpone that.

I found a few things that broke though and that should be fixed imho. Can you take another look at it and maybe also click through the application on your machine? Thanks a lot in advance!

PR #463, opened 2024-12-11
Small type change to improve i18n API

opyh:
This simplifies the typing for `translatedStringFromObject()` users a bit.

Using `undefined` as return value is better than `null`. There is no need to *explicitly* return "There is no value!", which simplifies type annotations for all consumers.


kriskbx (approved):
Yeah, that makes a lot of sense. Thanks 🫶

PR #464, opened 2024-12-11
Add theme (incl. dark mode) support to category/a11y filters

opyh:
This just uses the correct CSS colors for the filtering UI, improving its look'n'feel in both bright and dark modes.


kriskbx (approved):
Nice, thanks a lot for this ❤️ Great work, LGTM!

**Non-Blocker:** The current design is quite good, but it doesn't 100% fit to the rest of radix-ui. What's your opinion, should we replace the buttons at a later stage with radix components?

opyh:
Yes, we should absolutely find a more fitting design for these sooner or later!

PR #465, opened 2024-12-11
Radix-related fixes, improved main menu + OSM sign-in

opyh:
- Fix change requests from https://github.com/sozialhelden/wheelmap-frontend/pull/459
- Fix <kbd>L</kbd> hotkey (this key unlocks the OSM sign-in menu item)
- Refactor main menu and sign-in with OSM to use Radix components

Screenshot of the new Radix UI-based main menu


src/components/App/MainMenu/SessionMenuItem.tsx
```diff
@@ -0,0 +1,69 @@
+import { signIn, signOut, useSession } from "next-auth/react";
```
kriskbx, line 1: I really really love how this section turned out. It looks amazing. Great work 🫶 

A man, probably from South-America, sitting in front of the camera, saying "Mucho mucho joy"

kriskbx (changes requested):
Great work, thanks a lot for your amazing PR ❤️ I really like the direction, the interface is going now. 😻 

**Bug:** One thing that needs to be fixed though: after opening the dropdown menu either through the mouse or by focusing it and using enter/space I cannot navigate the items with the arrow keys. After pressing an arrow key, the error "TypeError: Cannot read properties of null (reading 'focus')" gets thrown. Can you take a look at this?

**Discussion/Suggestion:** I like that all those navigation items got moved into a dropdown menu, it makes the interface more clear and removes distractions. However, this also means that users will have a harder time finding things. This can have some negative implications especially when we want them to find some things very easily. For example when we want them to participate, e.g. the button to "add a new place". I suggest to move the "add a new place" button out of the dropdown menu again right next to menu button and also make sure it visually pops out more, e.g. by using the color "gray" for the menu button and the primary color for the "add a new place" button. Maybe we could also use an icon instead the text "add a new place". I don't know, but worth trying out. I'm curious on your opinion @opyh and @Mayaryin, what do you think? Does that make sense? Do you have other ideas?

biome.json
```diff
@@ -14,7 +14,7 @@
 	},
 	"formatter": {
 		"enabled": true,
-		"indentStyle": "tab"
+		"indentStyle": "space"
```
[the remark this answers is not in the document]
opyh: Makes sense! I could create a new PR for this behavior.

src/components/App/MainMenu/MainMenu.tsx
```diff
@@ -1,315 +1,107 @@
...
-
-        <div id="main-menu" role="menu">
-          <AppLinks />
-        </div>
-
-        {closeButton}
-      </StyledNav>
-    </Card>
-	);
+        <VectorImage
```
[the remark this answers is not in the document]
opyh: Absolutely. We need logos in 2 variants for this to work. Will create an extra PR for this.

src/components/App/MainMenu/MainMenu.tsx
```diff
@@ -1,315 +1,107 @@
...
+
+  return (
+    <Theme radius="none">
+      <StyledCard variant="surface">
+        <Inset>
+          <Theme radius={radius}>
+            <Flex justify={"between"} align="center" p="2">
+              <Flex align="center" gap="4">
+                {logoHomeLink}
+                <Text as="div">{claimString}</Text>
```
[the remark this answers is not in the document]
opyh: You're right, this shouldn't even be visible on mobiles!

biome.json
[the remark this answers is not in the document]
kriskbx: I actually wanted to express my feeling of joy by posting that gif. I like the change. But yeah, let's do another PR. Fine for me 😄

src/components/App/MainMenu/link-types/JoinedEventLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { data: joinedMappingEvent, isValidating } = useCurrentMappingEvent();
+
+  const href = joinedMappingEvent
+    ? `/events/${joinedMappingEvent._id}`
+    : "/events";
+
+  const label = joinedMappingEvent ? joinedMappingEvent.name : props.label;
+  const children = isValidating ? <Spinner /> : <Text>{label}</Text>;
+  const router = useRouter();
+  // XXX: This should be an <AppStateAwareLink>, but this would break keyboard navigation:
```
opyh, line 23: This is the fix for the keyboard-related JS crash you mentioned.

src/components/App/MapLayout.tsx
```diff
@@ -50,34 +52,36 @@ export default function MapLayout({
...
 
-  const router = useRouter()
-  const { pathname } = router
-  React.useEffect(() => {
-    setIsMenuOpen(false)
-  }, [pathname])
+  const { toggleExpertMode } = useExpertMode()
+  const expertModeHotkeys: HotkeyConfig[] = React.useMemo(() => [
+    {
+      combo: 'mod+e',
```
opyh, line 64: Switched this to <kbd>Cmd</kbd>+<kbd>E</kbd> to avoid accidental activation.

src/components/App/MapLayout.tsx
```diff
@@ -50,34 +52,36 @@ export default function MapLayout({
 }) {
   const app = React.useContext(AppContext)
   const { clientSideConfiguration } = app || {}
-  const [isMenuOpen, setIsMenuOpen] = React.useState(false)
   const firstStart = isFirstStart()
-  const toggleMainMenu = React.useCallback((newValue?: boolean) => {
-    setIsMenuOpen(typeof newValue === 'boolean' ? newValue : !isMenuOpen)
-  }, [isMenuOpen])
+  const router = useRouter()
+  const isOnboardingVisible = firstStart || router.pathname === '/onboarding'
```
opyh, line 57: This is a part of the fix to get the missing onboarding dialog back when clicking the logo in the upper left.

src/components/CombinedFeaturePanel/PlaceLayout.tsx
```diff
@@ -30,11 +30,11 @@ export default function PlaceLayout({
 }) {
   return (
     <Toolbar>
-      <Link href={closeUrl ?? '/'}>
+      <AppStateLink href={closeUrl ?? '/'}>
```
opyh, line 33: Keeps the map state when closing the place details.

src/lib/fetchers/ac/refactor-this/fetchMappingEvent.ts
```diff
@@ -1,10 +1,10 @@
-import { MappingEvent } from '../../../model/ac/MappingEvent';
+import type { IImage } from '../../../model/ac/Image';
 import useDocumentSWR from '../useDocumentSWR'
 import { useMemo } from 'react'
 
 export function useMappingEvent(_id: string) {
-  const response = useDocumentSWR<MappingEvent>({
-    collectionName: 'MappingEvents',
+  const response = useDocumentSWR({
```
opyh, line 6: The API here wasn't correctly used before.

src/lib/util/useViewportSize.ts
```diff
@@ -1,34 +1,32 @@
...
+import { useState } from 'react'
+import { useIsomorphicLayoutEffect } from '../../components/shared/useIsomorphicLayoutEffect'
 
-const getBreakpointSize = (width: number) => (width <= 512 ? 'small' : 'big')
 
 /**
  * Determine the viewport size depending on the inner sizing of the window, useful
  * for handling layout choices depending on the size of the browser
  */
 export const useWindowSize = () => {
```
opyh, line 9: The changes here fix hydration errors and make server-side only renderings use 1024x768 as fallback window size.

src/components/App/MainMenu/link-types/SessionElement.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Button color="red" onClick={handleSignOut}>
+        {t`Sign out`}
+      </Button>
+    </Flex>
+  );
+}
+
+function NonAuthenticatedMenuContent() {
+  const handleSignIn = React.useCallback(() => signIn("osm"), []);
+  return <Button onClick={handleSignIn}>{t`Sign in`}</Button>;
```
opyh, line 45: Shortened link as simple fix for the too-long-text issue.

src/components/App/MainMenu/MainMenuLinks.tsx
```diff
@@ -0,0 +1,63 @@
...
+    [linksInToolbar],
+  );
+
+  const dropdownMenuButton = menuLinkElements.length > 0 && (
+    <Theme radius="small">
+      <DropdownMenu.Root onOpenChange={setIsOpen}>
+        <DropdownMenu.Trigger>
+          <Button
+            variant="solid"
+            color="gray"
```
opyh, line 41: New styling 💅

opyh:
Implemented the improvements you wished for (and hopefully some more!)

src/components/App/MainMenu/link-types/JoinedEventLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+  // A bug in Radix UI (?) appears to not update the internal `ref` correctly.
```
kriskbx, line 24: **Non-Blocker:** Maybe the AppStateLink needs to use forwardRef in order to be able to be used with ref? Idk, my knowledge of modern react is still a bit limited, but that came to mind. Nothing that we need to solve in this PR.

src/components/Map/filterLayers.ts
```diff
@@ -164,7 +164,7 @@ export function filterLayers(
       layout: omit(localizedLayer.layout, 'line-z-offset'),
       paint: omit(localizedLayer.paint, 'line-occlusion-opacity'),
     }
-    console.log(enhancedLayer)
```
kriskbx, line 167: Thank you 😄

kriskbx (changes requested):
Hey, thanks again for your amazing work all those nice changes ❤️  Unfortunately I still found some issues. 😭 

The "Events" Link is not a link but a span-tag. It also takes a few seconds after clicking before something happens without any immediate feedback. This is not accessible and causes other issues down the line. I would consider this a blocker, that needs to be changed before merging this PR.
image

**Non-Blocker:** The dropdown menu doesn't behave as expected. When opening the menu and pressing the down arrow, I would expect that the first item is highlighted. Unfortunately, that doesn't work. I can press the up-arrow and then the last item in the list is highlighted and I can navigate the list. This is a weird behaviour. It's still somewhat accessible and should not be blocking this PR, but something we should fix in the near future.
keyboard-nav-issue

src/components/App/MainMenu/link-types/JoinedEventLink.tsx
opyh: That was it! I was 100% sure I had tried this, but missed it.

opyh:
Thanks for your thorough testing! I've moved the sign-in button out of the menu, added the missing `forwardRef()` wrapper, and arrow keys appear to work as expected now.

Could you have another look?

kriskbx (approved):
Thanks again for your amazing work. LGTM 🫶

PR #466, opened 2024-12-11
feat: radix image gallery

kriskbx:
This is the first of a set of PRs for improving and fixing the image/photo gallery feature. I'm splitting these in order to speed up code reviews. This PR focuses on the actual gallery and does not include changes to the "image upload" and "report image" feature.

I rewrote the image gallery using components from the newly introduced radix-ui. I also changed how it handles state and how it's split into different components. I think the new code makes it easier to understand and thus maintain in the future.

There are a still few open things:

1. There are some "flashes" caused by the application rerendering stuff when the route changes, e.g. when opening or closing the dialog. The rerendering doesn't occur, when I temporarily disable the route changes. I think, there's something going on higher up in the component tree, but I couldn't find anything obvious. Due to my limited knowledge of the codebase, I probably need some help here.
2. We don't have alt text for images (yet). So, when navigating it by screen-reader I currently implemented to read out the id of the image. Not optimal, but at least we have an implementation to flip out, when generating alt-text with AI or something later on.

Other than that I added some comments where I thought it make sense, because it either deviates from the standard or is a little hard to understand (see useEffect hooks).

Looking forward to your feedback!


src/components/CombinedFeaturePanel/components/Gallery/Gallery.tsx
```diff
@@ -1,64 +1,185 @@
...
+  // opened the Dialog in the first place, so that's why this functionality is implemented a
+  // second time.
+  const galleryRef = useRef<HTMLElement>(null);
+  const handleOnCloseAutoFocus = (event: Event) => {
+    event.preventDefault();
+    galleryRef.current
+      ?.querySelector<HTMLAnchorElement>(`[data-image-id=${activeImage?._id}]`)
+      ?.focus();
+  };
+
```
Mayaryin, line 136: This does not seem to work for me (Firefox and Chrome). After leaving the full screen image, no element is focused on the map screen

Mayaryin (commented):
Cool work!

src/components/CombinedFeaturePanel/components/Gallery/Gallery.tsx
```diff
@@ -1,64 +1,162 @@
...
+        next();
+        break;
+      default:
+        break;
+    }
+  };
+
+  return (
+    <GalleryContext.Provider value={api}>
+      <Dialog.Root open={dialogOpen} onOpenChange={setDialogOpen}>
```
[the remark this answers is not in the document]
kriskbx: You're right. Good catch, changed it 🙏

src/components/CombinedFeaturePanel/components/Gallery/Gallery.tsx
kriskbx: Shame on me, I haven't tested in Chrome and for some weird reason it worked in Firefox at some time, even though it shouldn't have 😄 It should be fixed now!

Mayaryin: I tested the new commits on Chrome and it works now! Thank you :)

PR #469, opened 2024-12-21
To make test files readable

NinaEilers:
The file search-by-name.js contains two changes, that will help to wait longer times and therefore do not fail.


NinaEilers (commented):
There were in the other files, too: search-by-shop.js is not finished and highlighted-marker.js  contains one difference exchanging h1 by div and one adding a comma.

opyh (approved):
Nice! 👌

Let's talk about how to proceed with this next time we meet.

PR #471, opened 2024-12-30
fix: feature panel rerendering

kriskbx:
This fixes the re-rendering of the FeaturePanel and some flickering in the newly added image gallery. The fix for the image gallery that circumvents react-router should be a temporary one. As far as I understand, the best practise would be to have a single page for the FeaturePanel with a catch-all route that handles child-routes like image upload or the gallery dynamically instead of creating separate pages for them. I think, this is a pattern we should use in general when dealing with dialogs/modals that have their own route. I'm eager to hear your opinion on this issue @opyh @Mayaryin

### DIFF
--- a/src/components/CombinedFeaturePanel/components/Gallery/Gallery.tsx
+++ b/src/components/CombinedFeaturePanel/components/Gallery/Gallery.tsx
@@ -98,7 +98,14 @@ export const Gallery: FC<{
     if (activeImageId) goTo(activeImageId);
   }, [images, activeImageId]);
 
-  // Reset the url when closing the dialog
+  // Reset the url when closing the dialog. This is not using react router on purpose.
+  // When using react-router it will re-render the whole FeaturePanel, because it's a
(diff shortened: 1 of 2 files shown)


src/components/CombinedFeaturePanel/components/Gallery/Gallery.tsx
```diff
@@ -129,7 +136,7 @@ export const Gallery: FC<{
   const handleOnCloseAutoFocus = (event: Event) => {
     event.preventDefault();
     document
-      .querySelector<HTMLAnchorElement>(`[data-image-id=${activeImage?._id}]`)
+      .querySelector<HTMLAnchorElement>(`[data-image-id="${activeImage?._id}"]`)
```
kriskbx, line 139: I stumbled upon another bug when testing this and I thought it would be okay to fix it in this PR as well. When the activeImageId starts with a numeric character, quotes are needed.

opyh (approved):

PR #473, opened 2025-01-02
refactor: app context

kriskbx:
This refactors the _app.tsx and the app context. It removes the fetching of the actual app config from a11ycloud from the getInitialProps method to prevent it from being called on every navigation. It also refactors some request and response related code to utilities and cleans up the _app.tsx to only contain the code that actually belongs there.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209017527142668


src/lib/util/request.ts
```diff
@@ -0,0 +1,55 @@
...
+export function getRequestQuery(
+  request?: IncomingMessage,
+): ParsedQuery<string> {
+  return parseUrl(getRequestUrl(request)).query;
+}
+
+export function getRequestLanguageTags(request?: IncomingMessage): string[] {
+  if (request) {
+    return parseAcceptLanguageString(
+      request.headers?.["accept-language"] ?? "",
```
opyh, line 29: While most browsers send an `Accept-Language` header, our liveness probe does not, which likely is the reason why https://refactor-app-context.wheelmap.tech isn't reachable.

I guess this fails if there is no `accept-language` header. I'd suggest to fall back to 'English (US)' like this (or using the central constant):

```suggestion
      request.headers?.["accept-language"] ?? "en-US",
```

…or to simply return `'en-US'` in the function if the header is missing.

src/lib/util/request.ts
```diff
@@ -0,0 +1,55 @@
...
+  request?: IncomingMessage,
+): string | undefined {
+  const countryCode =
+    (request?.headers?.["cf-ipcountry"] as string) ||
+    (request?.headers?.["x-country-code"] as string);
+  if (countryCode) {
+    return countryCode;
+  }
+  return (
+    getRequestILanguageTags(request).find((tag) => !!tag.region)?.region ?? ""
```
opyh, line 53: As `""` is no valid/parseable country code, I'd suggest to assert a valid language tag / throw an exception if the request has no tag (the change above would ensure it always *does* has a tag 😄).

src/pages/_app.tsx
```diff
@@ -1,199 +1,134 @@
...
+                      <LanguageCodeContextProvider languageTags={languageTags}>
+                        <AppContextProvider>
+                          {getLayout(<Component />)}
+                        </AppContextProvider>
+                      </LanguageCodeContextProvider>
+                    </CountryContext.Provider>
+                  </UserAgentContextProvider>
+                </HostnameContextProvider>
+              </EnvContextProvider>
+            </SWRConfigProvider>
```
opyh, line 85: While I personally liked the previous nesting function, this is easier to understand (and adds æsthetics of an arrow pointing forward), so I still like it 🤝

opyh (changes requested):
I like, thank you for this! Just one very small change request, otherwise this looks fabulous 🙏

src/pages/_app.tsx
kriskbx: ➡️

PR #482, opened 2025-01-21
Add tag name to request body for refresh endpoint

Mayaryin:
Added tag name to request body to make it work with the changes in the osm api. Only works with the corresponding branch (same name) in the osm-api repo.

### DIFF
--- a/src/lib/fetchers/callBackendToUpdateInhouseDb.ts
+++ b/src/lib/fetchers/callBackendToUpdateInhouseDb.ts
@@ -4,10 +4,12 @@ export async function callBackendToUpdateInhouseDb({
   baseUrl,
   osmId,
   osmType,
+  tagName,
 }: {
   baseUrl: string;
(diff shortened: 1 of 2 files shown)


src/lib/fetchers/callBackendToUpdateInhouseDb.ts
```diff
@@ -18,6 +20,7 @@ export async function callBackendToUpdateInhouseDb({
     },
     method: 'POST',
     mode: 'cors',
+    body: tagName
```
opyh, line 23: suggestion (if-minor): I'd prefer this to be in a query parameter or in the URL (like `${baseUrl}/legacy/api/${osmType}/${osmIdAsNumber}/refresh/${tagName}`), but as this is just for a test and we're going to have a v3 API after all our experiments, I'd be totally okay with this API for now :)

opyh (approved):

PR #489, opened 2025-02-06
Remove unused code, tidy up linter & typescript errors

opyh:
A photo of a wet raccoon

- Reduces number of Biome linting errors to zero
- Fixes a lot of TypeScript errors that popped up on the way, mostly manually
- Removes NPM dependencies that are not needed anymore
- Removes unreachable code (its kept in the git history)

This consists mostly of automated changes. Reviewing superficially is okay, I've mentioned code that could get a more thorough review.

This consists – mostly – of automated changes to the code base (using Biome's `--fix` argument). We agreed that this PR might introduce new bugs that we'll catch when cleaning up the release, but will remove more bugs by letting us start with 0 linting errors and less TypeScript errors.

The deployment doesn't go through, but this is already fixed in the [big merge PR where this code is supposed to end up](https://github.com/sozialhelden/wheelmap-frontend/pull/490).

<details><summary>🧼 I have cleaned up my code</summary>
  
- [x] I have removed dependencies that were just for testing.
- [x] I have removed debug logging.
- [x] My code does not generate new warnings.
- [x] My code does not depend on new vulnerable packages.
- [x] The commit messages are precise and make sense (rebase the PR with `--interactive` if applicable, keeping commits in sensible chunks if possible).
</details>

<details><summary>🔍 I have performed a self-review of my code</summary>
  
- [x] My code is self-documenting or has links to necessary documentation.
- [x] New function and variables names can be understood by new developers with basic project knowledge.
</details> 

<details><summary>✨ I have created a nice pull request</summary>
  
- [x] It has a clear title.
- [x] It follows the template, has a clear description and testing instructions if needed.
- [x] It targets the right branch.
- [x] I removed not applicable sections of the PR template.
- [x] [optional] I added a GIF of my favorite animal to the PR description to lighten the mood of my colleagues.
</details>

PR #497, opened 2025-02-12
feat: need picker and new category filtering

kriskbx:
##  🖊️ Description
**💡 What? Why? How?**

elephant-hello

This introduces the new need-picker and the new category filter menu.

**📎 Related Ticket**

https://app.asana.com/0/1208856252518002/1209215588864273

**🔗 PR Dependencies**

#489 

## ⚠️ Creation checklist

**Before creating this merge request, go over the following checklist (click to expand). 
Remove any items that are not applicable.**

  - [x] A new E2e playwright test covers this feature / A new test that reproduces the bug passes now.
  - [ ] The feature deployment works.
  - [x] The automated tests are passing.
  - [x] I have manually tested this feature
    - [x] on mobile
    - [x] by using keyboard-only navigation
    - [ ] with a screen reader (VoiceOver is fine)
    - [x] in Chrome
    - [x] in Firefox
    - [x] in Safari
</details>

- [x] I have removed dependencies that were just for testing.
- [x] I have removed debug logging.
- [x] My code does not generate new warnings.
- [ ] My code does not depend on new vulnerable packages.
</details>

- [x] My code is self-documenting or has links to necessary documentation.
- [x] New function and variables names can be understood by new developers with basic project knowledge.
</details> 

- [x] It has a clear title.
- [x] [optional] I added a GIF of my favorite animal to the PR description to lighten the mood of my colleagues.
</details> 

<details><summary>📝 I updated the documentation</summary>
  
- [ ] I updated the documentation in this repository.
- [ ] I updated the [tech manual](https://manual.i.wheelmap.tech/).
- [ ] I updated the manual testing plan of the app so that it includes a testing flow for this feature.
</details> 

## 🔍 Reviewing

## 🍸 Optional: Installation instructions
Run `npm ci` before testing locally.


.run/Template Playwright.run.xml
kriskbx, line 1: This is used by JetBrains IDEs and configures the "Test Automate" plugin for Playwright, so it should™️ work out of the box without the need to configure it manually.

e2e/lib/base-url.ts
kriskbx, line 1: If the `baseUrl` is configured correctly in playwright, you don't need it in the tests, you can just navigate to `"/"`.

e2e/search-by-name.spec.ts
kriskbx, line 1: There's a new and more in-depth test for this in the search domain.

package.json
```diff
@@ -143,9 +144,6 @@
     "> 0.25%",
     "not dead"
   ],
-  "engines": {
-    "node": "22"
-  },
```
kriskbx, line 148: Single source of truth should be the `.nvmrc` 🙂

package.json
```diff
@@ -111,7 +112,7 @@
     "push-translations": "rm -rf src/.next/ && npx ttag update public/i18n/translations.pot ./src && tx push -s && echo \"Translators can now start translating new strings:\\n\\nhttps://www.transifex.com/sozialhelden/wheelmap-react-frontend/translate/#de/translationspot/335454735?q=translated%3Ano\"",
     "pull-translations": "tx pull -af && node scripts/parseGettextFiles.js",
     "retranslate": "npm run build && npm run pull-translations && NEW_SOURCE_PO='./src/public/i18n/en_US.txt' retranslate/run.sh",
-    "compile-svg-directory": "echo \"Compiling SVGs in ${basedir}, this can take up to a few minutes…\"; (for file in ${basedir}/*.svg; do node node_modules/@svgr/cli/bin/svgr --icon $file > ${file/.svg/.tsx}; sed -i '' -e 's/= props/= (props: React.SVGAttributes<SVGElement>)/g' ${file/.svg/.tsx}; echo \"export { default as `basename $file .svg` } from './`basename $file .svg`';\";  done;) > ${basedir}/index.ts",
+    "compile-svg-directory": "npx @svgr/cli --out-dir ${basedir} -- ${basedir}",
```
kriskbx, line 115: This changes the svgr build prozess, as it now includes the functionality to create a index.ts file. This could introduce issues when recompiling other icons, I would like to tackle this in a separate PR.

playwright.config.ts
```diff
@@ -4,22 +4,24 @@ import { defineConfig, devices } from "@playwright/test";
...
 
 const baseURL = process.env.CI_TEST_DEPLOYMENT_BASE_URL;
 
 /**
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
-  testDir: "./e2e",
+  testDir: "./",
+  testMatch: "**/*.e2e-spec.ts",
```
kriskbx, line 16: I was thinking `e2e-spec` for Playwright tests and `spec` for jest unit-tests.

playwright.config.ts
```diff
@@ -4,22 +4,24 @@ import { defineConfig, devices } from "@playwright/test";
...
   /* Run tests in files in parallel */
   fullyParallel: !process.env.CI,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 3 : undefined,
+  workers: process.env.CI ? 3 : 1,
```
kriskbx, line 24: Multiple workers locally can run into issues when targeting the dev-server with hot module reloading and stuff.

tests/e2e/setup/test-fixture.ts
kriskbx, line 1: For things like code coverage in the future, we need to instrument and extend the test even more. That's why I thought I make it generic right now instead of axe-specific.

src/domains/needs/components/NeedsDropdownSection.tsx
```diff
@@ -0,0 +1,132 @@
...
+  font-size: .95rem;
+`;
+const HelpText = styled(Text)<{ $isVisible: boolean }>`
+  opacity: ${({ $isVisible }) => ($isVisible ? "1" : "0")};
+  max-height: ${({ $isVisible }) => ($isVisible ? "10rem" : "0")};
+  margin: ${({ $isVisible }) => ($isVisible ? ".25rem 0 .75rem" : "0")};
+  transition: all 300ms ease;
+`;
+
+export const NeedsDropdownSection = forwardRef(function NeedsDropdownSection(
```
Mayaryin, line 36: Thought: I dont think the combination of "No wheelchair info yet" and "Fully wheelchair accessible toilet" is possible. Maybe the toilets section could be hidden when "No wheelchair info yet" is selected. Or the no info option could be pulled out of the needs section and become a separate filter, since it also not really describes a need in the narrow sense.

Mayaryin (commented):
Really really great work! I can see that a lot of thougth and work was put into this 👏👏
I left some comments mostly from the user's perspective 😊

src/domains/needs/components/icons/mobility/PartiallyWheelchairAccessibleCombinationIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+      >
+        <path
+          d="M18 10C18 14.9706 13.9706 19 9 19C4.02944 19 0 14.9706 0 10C0 5.02944 4.02944 1 9 1C13.9706 1 18 5.02944 18 10Z"
+          fill="#46A758"
+        />
+        <path d="M17 0L25.6603 5V15L17 20L8.33975 15V5L17 0Z" fill="#FFBA18" />
+      </svg>
+    );
+  },
+);
```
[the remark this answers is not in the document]
kriskbx: I've adopted this from the current Wheelmap. If I had to guess, I'd say the intention is that choosing this setting will show fully accessible places as well as partially accessible ones (which it does). I'm also not 100% happy with this, but I kept the old icons for a few reasons:
1. I don't wanted to change too much at once, to minimize the confusion of existing users.
2. When we introduce more categories other than toilets and mobility we also need to revamp the iconography, probably changing these icons completely.
So, in the end I decided to keep it just like it is now.

src/domains/needs/components/NeedsDropdownButtonSection.tsx
```diff
@@ -0,0 +1,38 @@
...
+  function NeedsDropdownButtonSection(
+    {
+      onSaveButtonClick,
+      onResetButtonClick,
+      ...props
+    }: { onSaveButtonClick: () => void; onResetButtonClick: () => void },
+    ref: Ref<HTMLDivElement>,
+  ) {
+    return (
+      <Theme {...props} ref={ref} asChild radius="medium">
```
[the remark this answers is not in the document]
kriskbx: Yes, you're right. I deviated deliberately from the default border radius to better match the design, but we should use a consistent radius here, I will change that.

Good point: The variant and color choice for primary, secondary and tertiary actions should also be consistent. I personally favor a hierarchy-first approach, using the accent color only for the primary action. But it would be a valid approach to use it for the other levels as well. I made an overview with a few choices that could work. What do you think?

image

I disagree regarding the usage of icons rather than text. When I recall correctly, the accessibility recommendation is to use text or a combination of text and icons whenever you can. It is less error-prone than icon-only buttons and usually better understandable. What does a tick or a cross mean in this context? "Save" and "Cancel" is much more precise than icons, that tend to be interpreted differently by different user-groups.

src/domains/needs/components/NeedsDropdownButtonSection.tsx
```diff
@@ -0,0 +1,38 @@
...
+            {t`Save`}
+          </Button>
+          <Button
+            size="2"
+            variant="soft"
+            color="gray"
+            onClick={onResetButtonClick}
+          >
+            {t`Cancel`}
+          </Button>
```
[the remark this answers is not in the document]
kriskbx: I also thought about that, but I deliberately designed it without a reset button. The idea is: We want the user to make a selection, because it is very valuable information. If you select "no-needs" in a category, we know that this is something we don't need to take into account. So we can adapt the interface or other things behind the scenes to increase the usability in other places. If we don't have a selection, because the user can just reset the form, we don't know if they missed it or if they don't actually have this need. For example, when the user chooses "no mobility needs", we could preconfigure the app to not show surfaces (this could be toggled on again in the settings later).

What we could introduce is a "I have no needs" button instead of a reset button, that sets every category to "no-needs". Alternatively, and this is my favorite solution, we could save the selection in localStorage and in the native app somewhere in the file system, so the user only has to make a selection once.

src/domains/needs/components/NeedsDropdownSection.tsx
kriskbx: As far as I know, technically speaking, there can be places with an accessible toilet that have no accessibility rating themself. E.g. when multiple venues share the same toilets like in a shopping center. But I'm not 100% sure about that. This is something @opyh probably knows best.

> Or the no info option could be pulled out of the needs section and become a separate filter, since it also not really describes a need in the narrow sense.

You could argue that people attending mapping events have the need to find places without accessibility information. You could also argue that "not wheelchair accessible" is also not a need in the narrow sense, except maybe for people that want to improve accessibility in their area.

But I agree with you, I have the same feeling that the current implementation is not 100% ideal. But I don't think introducing rules to hide sections based on a previous selection or pulling out options and introducing a third filter-level (we already got two) are solutions to this problem that improve usability and that we should pursue. I'm eager to hear what others and later our users think about it though. And I would be very happy to build it differently when there's a good argument for it.

package.json
opyh: Right, and this is more relevant for packages, not for an app 🤝

src/domains/needs/components/icons/mobility/PartiallyWheelchairAccessibleCombinationIcon.tsx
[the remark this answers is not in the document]
opyh: This was yellow before a user test. Several people gave feedback that they did not understand why choosing a yellow marker would also display green markers. Neither of both solutions would be completely intuitive without some learning, and both solutions would lead to "I don't fully understand this" at first sight, so we could as well just display only a yellow marker.

src/domains/needs/components/NeedsDropdownButtonSection.tsx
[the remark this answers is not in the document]
opyh: Re radiuses: it makes sense to adapt them to their context. I don't know if Radix' design system does a good job in this regard, we should check this :)

An image showing how to create nested elements with round corners that aesthetically fit each other.

Primary/secondary/tertiary: Clear preference for variant 6 🙋

src/domains/needs/components/NeedsDropdownButtonSection.tsx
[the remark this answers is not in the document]
opyh: Yeah, saving this makes sense. localStorage functionality works in the native app as well :) I'd go without an explicit extra button as well, but look into this again after a test with 'real' users - might be that we do need a reset / 'show all' button eventually.

src/domains/needs/components/icons/mobility/PartiallyWheelchairAccessibleCombinationIcon.tsx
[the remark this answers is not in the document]
kriskbx: So, it seems like is not that big of an issue right now. So, let's go with the current solution that has been learned by users and revisit this, when we introduce new "needs" and require different iconography.

src/domains/needs/components/NeedsDropdownButtonSection.tsx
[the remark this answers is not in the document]
kriskbx: I added logic to save the selection in localStorage.

src/domains/needs/components/NeedsDropdownButtonSection.tsx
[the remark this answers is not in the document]
kriskbx: Clear preference for variant 2 here 👋

src/domains/needs/components/NeedsPicker.tsx
```diff
@@ -0,0 +1,146 @@
...
+`;
+
+export default function NeedsPicker() {
+  const { selection: globalSelection, setSelection: setGlobalSelection } =
+    useNeeds();
+  const [selection, setSelection] = useState<NeedSelection>(globalSelection);
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  const { sections, highlightedSection, isGivenOrNextSectionHighlighted } =
+    useNeedsHighlighterSections({ needs: selection });
```
opyh, line 68: I really like this clarity! 👏

src/domains/needs/components/icons/mobility/NoDataIcon.tsx
```diff
@@ -0,0 +1,18 @@
+import { type MutableRefObject, forwardRef } from "react";
+
+export const NoDataIcon = forwardRef(function NoDataIcon(props, ref) {
```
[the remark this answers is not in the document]
kriskbx: image

I played a little with this idea. Mixing no-needs with a need selection doesn't feel right. I guess there could be some kind of an empty state when "no-needs" is selected in all categories. To be honest, I don't think if this information is very useful and my gut tells me to keep it how it is. But I'm not sure. What do you think @Mayaryin @opyh ?

src/domains/needs/needs.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+// we're using const assertions in order to automatically generate types
+// from the settings. but in order to make sure the settings above also
+// satisfy their own interface, we have a validatedSettings variable typed
+// with the aforementioned interface. if something is off with the settings
+// above, typescript will show an error on this validatedSettings variable
+// instead, even though it's not in use anywhere. this is not ideal, but it
+// ensures type-safety and allows for auto-type magic.
+const validatedSettings: Record<
```
opyh, line 101: quip: I get why this is there. I'm wondering if we can figure out a better solution that fulfils the same requirements. :)

src/domains/needs/tests/needs-picker.e2e-spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    await getDropdown(page).getByRole("button", { name: "Cancel" }).click();
+
+    await assertDropdownIsNotVisible(page);
+  });
+
+  test("it shows all configured needs", async ({ page }) => {
+    await openDropdown(page);
+
+    await forEachNeedCategory(async ({ needs }) => {
+      await forEachNeed(needs, async ({ label }) => {
```
opyh, line 73: Much love for this 🥹

> Nobody: How much effort could someone put into a speedier test?
> You: Yes.

src/domains/search/components/hooks/useHighlightSearchResults.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function useHighlightSearchResults({
+  searchResults,
+  searchTerm,
+}: { searchTerm: string; searchResults?: EnrichedSearchResult[] }) {
+  const [highlightedIndex, setHighlightedIndex] = useState<number>(-1);
+  const searchResultsContainer = useRef<HTMLElement | undefined>();
+
+  const getHighlightedDomElement = () => {
+    // TODO: refactor this to use refs instead of querySelector
```
opyh, line 12: Later… :)

src/domains/search/functions/data-mapping.ts
```diff
@@ -49,30 +28,69 @@ export const makeDisplayDataFromPhotonResult = (
...
+
+export function mapOsmType(feature: PhotonResultFeature) {
+  return typeMap[feature.properties.osm_type || "N"] || "node";
+}
+
+const collectionMap = {
+  elevator: "elevators",
+  "highway:elevator": "elevators",
+  building: "buildings",
+  "place:house": "entrances_or_exits",
```
opyh, line 59: quip (non-blocking): From my understanding `place:house` is a different entity on OSM than our `entrances_and_exits` table – the nouse numbers are their own nodes so you can put them at a specific position on a map, usually in the center of a building, but we can figure this out later.

src/domains/search/tests/search-input.e2e-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    await page.waitForURL("**/amenities/node:3908141014**");
+  });
+
+  test("it shows an empty state if not results are found", async ({ page }) => {
+    test.slow();
+    await expect(
+      getSearchDropdown(page).getByText("No results found!"),
+    ).toHaveCount(0);
+
+    // if a place named like this exists it is well deserved to break our tests 🖖
```
opyh, line 65: 😆 First thought: [Boltzmann brain](https://en.wikipedia.org/wiki/Boltzmann_brain)

opyh (approved):
This is amazing. The clean semantics and thoughtful code make it easy to read despite its internal UI complexity. Very well done 💯

Nothing blocking, just a few suggestions/questions that we might as well clarify later.

src/domains/needs/components/icons/mobility/NoDataIcon.tsx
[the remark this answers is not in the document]
opyh: While I get the idea, I'd prefer context over consistency here. So I'd also not show a specific state-dependent extra label if the user basically selects "I don't care".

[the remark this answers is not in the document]
opyh: The same applies to the (2) label: in my opinion, if I select the "I have no needs" option two times, the button could simply stay in its initial state.

opyh:
Suggestion (non-blocking): I'd prefer a simpler wording for the "don't care" options, like

`I have no mobility needs` -> `I don't care`
`I have no toilet needs` -> `I don't care`

– "I have no toilet needs" sounds a bit like the alternative is that the user needs a toilet right now.

src/domains/needs/components/NeedsPicker.tsx
```diff
@@ -0,0 +1,146 @@
...
+import type { NeedSelection } from "~/domains/needs/needs";
+import { useNeeds } from "~/domains/needs/hooks/useNeeds";
+import { NeedsButton } from "./NeedsButton";
+import { NeedsDropdownHighlighter } from "./NeedsDropdownHighlighter";
+import { NeedsDropdownHighlighterSectionContainer } from "./NeedsDropdownHighlighterSectionContainer";
+import { NeedsDropdownSection } from "./NeedsDropdownSection";
+import { useNeedsHighlighterSections } from "~/domains/needs/components/hooks/useNeedsHighlighterSections";
+
+const NeedsDialogOverlay = styled(DialogPrimitive.Overlay)`
+  animation: showOverlay 400ms ease-out;
```
opyh, line 15:
```suggestion
  animation: showOverlay 400ms cubic-bezier( 0.23, 0.72, 0.3, 1 );
```

src/domains/needs/components/NeedsPicker.tsx
```diff
@@ -0,0 +1,146 @@
...
+      opacity: 0;
+    }
+    to {
+      opacity: 1;
+    }
+  }
+`;
+const NeedsDialogContent = styled(DialogPrimitive.Content)`
+  box-sizing: border-box;
+  animation: showContent 400ms ease-out;
```
opyh, line 31:
```suggestion
  animation: showContent 400ms cubic-bezier( 0.23, 0.72, 0.3, 1 );
```

opyh (approved):

src/components/CombinedFeaturePanel/components/Gallery/GalleryFullscreenItem.tsx
```diff
@@ -52,7 +52,7 @@ const ImageNav = styled.nav`
   transform: translateY(-50%);
 `;
 const ImageNavList = styled.ul`
-  list-style: none;
+  list-style-type: "";
```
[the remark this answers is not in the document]
kriskbx: This was done deliberately to keep the implicit list role and it's semantics for screen-readers. Setting it to none will lead to screen readers treating it as a normal container.

src/components/icons/mainCategories/undefined.tsx
```diff
@@ -1,17 +1,21 @@
...
+    height="1em"
+    viewBox="0 0 15 15"
+    aria-hidden="true"
+    ref={ref}
+    {...props}
+  >
+    <circle cx={7.5} cy={7.5} r={4.5} fill="currentColor" fillRule="evenodd" />
+  </svg>
+);
+const ForwardRef = forwardRef(SvgUndefined);
```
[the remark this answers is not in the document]
kriskbx: YES PLEASE!

src/domains/categories/categories.ts
```diff
@@ -0,0 +1,105 @@
...
+  Health,
+  Leisure,
+  MoneyPost,
+  PublicTransfer,
+  Shopping,
+  Sport,
+  Tourism,
+} from "~/components/icons/mainCategories";
+
+export const categories = {
```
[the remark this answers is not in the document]
kriskbx: See comment below!

src/domains/search/functions/data-mapping.ts
kriskbx: Yes, let's tackle this later. I only refactored this and didn't add or change anything in the process.

src/domains/needs/components/NeedsDropdownButtonSection.tsx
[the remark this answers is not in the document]
kriskbx: This is being handled in a separate PR: #501

src/domains/needs/components/icons/mobility/NoDataIcon.tsx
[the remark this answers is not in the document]
kriskbx: Let's test this with actual users and if it comes up, we can still implement this.

src/domains/needs/components/NeedsDropdownSection.tsx
kriskbx: I'm also resolving this for now, but I put it on my list to take a look at it later.

src/domains/categories/functions/cache.ts
```diff
@@ -38,16 +21,18 @@ export function getCategory(
 
 export function generateSynonymCache(categories: ACCategory[]): SynonymCache {
   const result: SynonymCache = new Map();
-  for (const category of categories) {
+  // biome-ignore lint/complexity/noForEach: <explanation>
+  categories.forEach((category) => {
```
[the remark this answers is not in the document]
kriskbx: I like the reminder that for of loops exists, but in this case, I prefer forEach as it is cleaner. Dunno if I would disable it completely.

PR #501, opened 2025-02-21
Chore/unify buttons

Mayaryin:
GIF

##  🖊️ Description
**💡 What? Why? How?**

We were lacking a unified look for standard buttons used throughout the app.
Following @kriskbx' ideas in our discussion on #497 , I propose 3 button styles for primary, secondary and tertiary buttons in this PR.
Secondary and tertiary buttons use high contrast setting.

[**📎 Related Ticket**](https://app.asana.com/0/1209303193509277/1209421365152962)

Depends on #497 

[**👀 Seak Peek**](https://chore-unify-buttons.wheelmap.tech/)

## ⚠️ Creation checklist

  - [x] The feature deployment works.
  - [x] The automated tests are passing.
  - [x] I have manually tested this feature
 
</details>

- [x] My code does not generate new warnings.
</details>

</details> 

- [x] It has a clear title.
- [x] [optional] I added a GIF of my favorite animal to the PR description to lighten the mood of my colleagues.
</details> 

## 🔍 Reviewing


src/components/shared/Buttons.tsx
```diff
@@ -0,0 +1,59 @@
...
+      variant="solid"
+      size={size}
+      onClick={onClick}
+      disabled={disabled}
+      loading={loading}
+    >
+      {text}
+    </Button>
+  );
+};
```
kriskbx, line 31: **issue:** Use forwardRef to enable reference to the underlying dom element.

The new button components cannot be used with refs, because they don't forward it. I know a few places in radix that rely on the ref-functionality and could break without it, so we should add it: https://18.react.dev/reference/react/forwardRef

src/components/shared/Buttons.tsx
```diff
@@ -0,0 +1,59 @@
...
+import type { Responsive } from "@radix-ui/themes/dist/cjs/props";
+import type React from "react";
+
+type Props = {
+  text: string;
+  onClick: () => void;
+  size?: Responsive<"3" | "1" | "2" | "4">;
+  disabled?: boolean;
+  loading?: boolean;
+};
```
kriskbx, line 11: **suggestion**: Use the original radix prop type and add your own extensions.

Please utilize the ButtonProps from @radix-ui/themes. This removes a lot of duplicated code and brings the benefit in combination with prop spreading (see comment below) to support all properties of radix-buttons.

```suggestion
type Props = { iconOnly?: boolean; } & ButtonProps;
```

src/components/shared/Buttons.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const PrimaryButton = ({
+  text,
+  onClick,
+  size = "3",
+  disabled = false,
+  loading = false,
+}: Props) => {
+  return (
+    <Button
```
kriskbx, line 21: **suggestion:** Add possibility to use IconButton.

There are 2 different buttons in radix: The normal Button for text (with or without an icon) and the IconButton for icon only buttons. We have a few icon-only buttons in the UI, so I suggest adding a property like `iconOnly` and switch between them:

```suggestion
  const Component = iconOnly ? IconButton : Button;
  return (
    <Component
```

src/components/shared/Buttons.tsx
```diff
@@ -0,0 +1,59 @@
...
+      {text}
+    </Button>
+  );
+};
+
+export const TertiaryButton = ({ text, onClick, size = "3" }: Props) => {
+  return (
+    <Button
+      color="gray"
+      variant="surface"
```
kriskbx, line 51: **suggestion (non-blocking):** Let's talk variants:

image

I favor option 2, Sebastian favors option 6.

I think option 2 is nice as it uses the accent color and even the tertiary button still looks like something you can click compared to the gray variants and because I like the looks of it. But you have a good point here: The surface variant looks more like a button and is probably better for accessibility than the ghost variant. But I think the surface variant is not distinctive enough from the soft variant, so if we want to go with a border here, I suggest using the outline variant instead of surface (this would be option 1 or 4).

kriskbx (changes requested):
Thanks a ton for your great PR 🫶 Amazing GIF and amazing work ❤️  

There are a few things that need to be changed in the wrapper components and we should also discuss about what variant we want to use.

src/components/shared/Buttons.tsx
Mayaryin: Great suggestion, thanks a lot!

src/components/shared/Buttons.tsx
Mayaryin: yes!

src/components/shared/Buttons.tsx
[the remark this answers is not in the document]
Mayaryin: Thanks a lot for this remark, it makes a lot more sense like this 💡

src/components/shared/Buttons.tsx
```diff
@@ -0,0 +1,59 @@
...
+type Props = {
+  text: string;
+  onClick: () => void;
+  size?: Responsive<"3" | "1" | "2" | "4">;
+  disabled?: boolean;
+  loading?: boolean;
+};
+
+export const PrimaryButton = ({
+  text,
```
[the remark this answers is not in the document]
Mayaryin: aaaah 💡

src/components/shared/Buttons.tsx
Mayaryin: i changed to option 1

src/components/CombinedFeaturePanel/components/ImageUpload/ImageUploadDropzone.tsx
```diff
@@ -75,13 +76,11 @@ export const ImageUploadDropzone: FC = () => {
         <TextForMobile as="p">
           <Strong>{t`Use the following button to select an image:`}</Strong>
         </TextForMobile>
-        <Button variant="soft" mt="2">{t`Select image`}</Button>
+        <SecondaryButton mt="2">{t`Select image`}</SecondaryButton>
```
kriskbx, line 79: **issue (non-blocking)**: These buttons have the same visual hierarchy now (both secondary). Before, there was a difference in hierarchy due to different colors. In order to establish a clear visual hierarchy again (the "select image button" is the one that should be pressed), I propose to use PrimaryButton either here or TertiaryButton for the back button.

src/components/shared/Buttons.tsx
kriskbx, line 1: **nit-pick (non-blocking):** I personally would prefer to have individual files for each Button component, but this way is also totally fine.

kriskbx (approved):
Thank you! Amazing work 🫶

src/components/shared/Buttons.tsx
Mayaryin: I was overthinking about just that question and then just went for one file, but I really do not have any preference :D

PR #569, opened 2025-08-12
Basic Screenreader Accessibility

Mayaryin:
gif

##  🖊️ Description
**💡 What? Why? How?**
This PR focuses on making the non-map parts of the app accessible for screenreader users. 

I added landmarks for the the search, need-picker, and the category filter. I would have preferred to make the latter two regions but voice over currently does not recognize regions as landmarks, so the only option was to make them navigation elements.

I refactored FeatureNameHeader into a smaller component FeatureHeader that is easier to understand and adapt but left the original component in the code base since it cointains additional functionality that we might want to use again later on. I also slightly refactored OpeningHoursValue visually.

I added a PageTitleContext in order to be able to change the page title depending on the POI to make the screenreader announce the page title correctly when changing the page. Somehow, this did not work reliably so I also set the focus on the place name header as soon as the side bar opens so that the place title is announced. The sidebar is also a live region now and when selecting a category and zooming in and out on the map it announces the sidebar content.

I also added headers to the feature details, a lang tag in the document head and aria-labels where they were missing and some other adjustments.

Voiceover didnt open radix' drop-down menu components since they dont react to a click event, so I added an extra key down event in their click handlers.

The general tab order in the app is still not optimal, since the map controls are coming before the search bar in the the tab order. Since experienced screen reader users like use the rotor and landmarks to explore a website this could be adressed later.

**➕ Additonal comments/questions**
**Missing:**
- The nested list that describes ACC places  is read out too verbosely. We need an extra concept for it since the data structure is relatively complex and many values need extra aggregated labels  [Ticket](https://app.asana.com/1/1200321573365931/project/1208833037492986/task/1211044733145345?focus=true)
- Markdown should be removed from value lables since it harms screenreader access (text is split and read out separately). Wheelmap pro currently has a bug that prevents access to UI vocuabulary tools
- Tests 🫠

[**📎 Related Ticket**](https://app.asana.com/1/1200321573365931/project/1208833037492986/task/1207796258426524?focus=true)

[**👀 Sneak Peek (Optional)**](https://feat-basic-screenreader.wheelmap.tech/)

## ⚠️ Creation checklist

**Before creating this merge request, go over the following checklist (click to expand). 
Remove any items that are not applicable.**

  - [ ] A new E2e playwright test covers this feature / A new test that reproduces the bug passes now.
  - [x] The feature deployment works.
  - [ ] The automated tests are passing.
  - [ ] I have manually tested this feature
    - [ ] on mobile
    - [ ] by using keyboard-only navigation
    - [x] with a screen reader (VoiceOver is fine)
    - [ ] in Chrome
    - [x] in Firefox
    - [ ] in Safari
</details>

- [x] I have removed debug logging.
- [ ] My code does not generate new warnings.
- [x] My code does not depend on new vulnerable packages.
</details>

- [x] My code is self-documenting or has links to necessary documentation.
- [x] New function and variables names can be understood by new developers with basic project knowledge.
</details> 

- [x] It has a clear title.
- [ ] [optional] I added a GIF of my favorite animal to the PR description to lighten the mood of my colleagues.
</details> 

<details><summary>📝 I updated the documentation</summary>
  
- [ ] I updated the documentation in this repository.
- [ ] I updated the [tech manual](https://manual.i.wheelmap.tech/).
- [ ] I updated the manual testing plan of the app so that it includes a testing flow for this feature.
</details> 

## 🔬 Optional: Testing instructions

Voice Over can be started and stopped pressing CMD+F5. The rotor can be accessed presing SHIFT+U and browsed by pressing the arrow keys.

## 🔍 Reviewing


src/components/sheet/Sheet.tsx
```diff
@@ -71,12 +72,13 @@ export function Sheet({
...
   isExpanded: externalIsExpanded,
   onIsExpandedChanged,
+  ...ariaProps
 }: {
   children: ReactNode;
   scrollStops?: number[];
   isExpanded?: boolean;
   onIsExpandedChanged?: (isExpanded: boolean) => void;
-}) {
+} & React.AriaAttributes) {
```
kriskbx: issue (non-blocking):

Essentially you're forwarding all properties to the underlying component but by typing it like this, typescript will complain if you use html attributes that are not aria-attributes. I suggest renaming `ariaProps` to `props` and typing it with `HTMLAttributes<"aside">` instead of `AriaAttributes`.

src/components/sheet/Sheet.tsx
```diff
@@ -135,6 +137,8 @@ export function Sheet({
     <SheetContainer
       $isExpanded={isExpanded}
       ref={container as Ref<HTMLDivElement>}
+      aria-label={t("Sheet")}
```
kriskbx: question:

What is the reason for labelling this component?

src/components/sidebar/Sidebar.tsx
```diff
@@ -88,15 +89,24 @@ export function Sidebar({
   isExpanded,
   onIsExpandedChange,
   hasPadding = true,
+  ...ariaProps
```
kriskbx: issue (non-blocking):

`props` instead of `ariaProps`, `HTMLAttributes<T>` instead of `AriaAttributes`.

src/components/sidebar/Sidebar.tsx
```diff
@@ -88,15 +89,24 @@ export function Sidebar({
...
   onIsExpandedChange: (value: boolean) => void;
-}) {
+} & React.AriaAttributes) {
   return (
-    <SidebarWrapper $isExpanded={isExpanded} $hasPadding={hasPadding}>
-      <SidebarToggleButton onClick={() => onIsExpandedChange(!isExpanded)}>
+    <SidebarWrapper
+      $isExpanded={isExpanded}
+      $hasPadding={hasPadding}
+      aria-label={t("Sidebar")}
```
kriskbx: question:

What is the reason for labeling this component?

src/hooks/usePageTitle.tsx
```diff
@@ -0,0 +1,31 @@
...
+  );
+}
+
+export function usePageTitle(newTitle?: string) {
+  const context = useContext(PageTitleContext);
+  if (!context) {
+    throw new Error("usePageTitle must be used inside <PageTitleProvider>");
+  }
+
+  if (newTitle !== undefined && newTitle !== context.title) {
```
kriskbx: suggestion (non-blocking):

I think this is not doing anything: `newTitle !== context.title`. From my understanding, if you call setState with the same value there will not be a rerender. It only rerenders, if the state actually changes.

src/hooks/usePageTitle.tsx
```diff
@@ -0,0 +1,31 @@
...
+    context.setTitle(newTitle);
+  }
+
+  return { title: context.title, setTitle: context.setTitle };
```
kriskbx: suggestion (blocking):

This API can be quite confusing. Either `usePageTitle("new title")` is used as a setter, then I would expect it to return either a boolean or the new page title. If it returns a `setTitle` function, I would recon it's not being used as a setter, but I need to use the `setTitle` function for setting the title. I have the feeling, that you wanted to support both things at the same time, but I think this makes the API less clear, so I suggest that you decide on one of them and either change the return or remove the parameter.

src/layouts/BaseMapLayout.tsx
```diff
@@ -74,14 +64,18 @@ export default function BaseMapLayout({
...
     <ErrorBoundary>
-      <HeadMetaTags />
-      <TopBar />
-      <Main $enablePaddingForSheet={isSheetMounted}>
-        <DynamicallyLoadedMap key="map" />
-        <div style={{ zIndex: 2000 }}>{children}</div>
-      </Main>
-      <ToastContainer />
-      {/*{isOnboardingVisible && <Onboarding />}*/}
+      <PageTitleProvider>
```
kriskbx: question:

All the context providers are located in `_app.tsx`. What is the reason to put it here?

src/modules/categories/components/CategoryFilter.tsx
```diff
@@ -1,12 +1,19 @@
...
 import styled from "styled-components";
 import { useCategoryFilter } from "~/modules/categories/contexts/CategoryFilterContext";
 import { getTopLevelCategoryList } from "~/modules/categories/utils/display";
 import { ChevronDown } from "lucide-react";
-import type { FC, SVGAttributes } from "react";
+import { type FC, type SVGAttributes, useRef } from "react";
+import { dispatchKeydownEvent } from "~/utils/dispatchKeydownEvent";
 
-const Container = styled.div`
+const Container = styled.nav`
```
kriskbx: issue (non-blocking):

This category filter is semantically speaking not a navigation as a navigation is a list of links. These are interactive buttons that trigger an action and thus are semantically a `menu`. I think this can be confusing, so I would suggest to change it. Also, `menu` is interpreted by all browsers as an unordered list, so it has the benefit of announcing how many items are in there.

src/modules/feature-panel/components/FeatureDetails.tsx
```diff
@@ -114,34 +123,51 @@ const FeatureDetails = ({
...
+              </div>
             )}
           </HeaderSection>
 
           <SectionsContainer $orderDesktop={3} $orderMobile={2}>
             {osmWheelchairInfo && (
               <Section>
+                <VisuallyHidden>
+                  <Heading as="h2">{t("Wheelchair section")}</Heading>
+                </VisuallyHidden>
```
kriskbx: praise:

Love those visually hidden headings for screen readers. Really n🍨

src/modules/feature-panel/components/FeatureDetails.tsx
```diff
@@ -68,14 +72,19 @@ const SectionsContainer = styled(Flex)<{
...
 }: Props) => {
   const feature = features[0];
-
   const map = useMap();
   const context = useContext(FeaturePanelContext);
   const { nestedTags } = useOsmTags(feature);
+  const { categoryName, placeName } = useFeatureLabel({ feature });
+
+  const pageTitle = placeName ? placeName : categoryName;
+  usePageTitle(pageTitle);
```
kriskbx: suggestion (blocking):

Instead of setting the page title deeply nested in a component, you could set it at the page level. (see comment above)

src/layouts/FeatureDetailsLayout.tsx
```diff
@@ -32,11 +34,18 @@ export const FeatureDetailsLayout = ({ children }: { children: ReactNode }) => {
...
   useEffect(() => {
     setIsExpanded(true);
   }, []);
 
+  useEffect(() => {
+    return () => {
+      setTitle(undefined);
+    };
+  }, [setTitle]);
+
```
kriskbx: suggestion (blocking):

Instead of doing this, you could set the page title for each individual page at the page level. So, in this case in the `pages/[placeType]/[id]/index.ts` file. Wenn you close the feature panel, it opens `pages/index.ts`, so there could be another `usePageTitle` that sets it to something else. Generally speaking, we want the usage of it to be consistent and predictable. Using an effect here in an layout file is confusing to me and quite unpredictable as it could clash with something else in the future.

src/needs-refactoring/components/App/HeadMetaTags.tsx
```diff
@@ -20,6 +23,20 @@ export default function HeadMetaTags() {
...
+
+  const categories = getCategories();
+  const categoryName =
+    isFilteringActive && category ? categories[category]?.name() : undefined;
+
+  const documentTitle = getProductTitle(
+    clientSideConfiguration,
+    specialisedTitle,
+    categoryName,
+  );
```
kriskbx: suggestion (blocking):

I suggest moving the all the page title logic to individual pages. Also see comments above. The category filtering is only needed for the `pages/index.ts` route, so it doesn't need to be here, but can live in that page. Best case: each page has to define it's own title.

src/modules/needs/components/NeedsPicker.tsx
```diff
@@ -93,60 +94,66 @@ export default function NeedsPicker() {
   }, [selection]);
 
   return (
-    <DialogPrimitive.Root open={isOpen} onOpenChange={setIsOpen}>
-      <DialogPrimitive.Trigger asChild>
-        <NeedsButton />
-      </DialogPrimitive.Trigger>
+    // using a <section> tag, role="region" or a <form> tag here did not make voice over present it as a landmark in the rotor
+    <nav aria-label={t("Needs Picker")}>
```
kriskbx: issue (blocking):

Just like the category filter, this is triggering a filter action and should not be a navigation as it is confusing that there's a form and not a list of links. When using a section with an aria-label it doesn't show up as a landmark, because it is a child of the header element which is itself a landmark. But it shows up under forms, which is fine I think. The only confusing thing right now: the need picker and the category filter, which are both filters, are in different places. One is in the header aka banner landmark and one is in the main section. I think this is okay for now, but we should find a way how to put them both in the same place or even combine them as they essentially do the same thing. Let's do this in a later stage.

src/modules/needs/components/NeedsPicker.tsx
```diff
@@ -93,60 +94,66 @@ export default function NeedsPicker() {
...
+        <DialogPrimitive.Trigger asChild>
+          <NeedsButton />
+        </DialogPrimitive.Trigger>
 
-      <DialogPrimitive.Portal>
-        <Theme>
-          <NeedsDialogOverlay />
-          <NeedsDialogContent>
-            <VisuallyHidden>
-              <DialogPrimitive.Title />
```
kriskbx, line 106: issue (blocking):

Removing this throws an error in the console:
Image

src/modules/needs/components/NeedsDropdownSection.tsx
```diff
@@ -92,14 +93,19 @@ export const NeedsDropdownSection = forwardRef(function NeedsDropdownSection(
...
+                        }
                       />
                       <Flex
                         as="span"
                         flexBasis="100%"
                         justify="between"
                         align="center"
                         gap="4"
+                        // to prevent voiceover from reading the label twice
+                        aria-hidden="true"
```
kriskbx: question:

The RadioGroup doesn't need a label, because there already is proper text in there. By adding the additional aria-label, you need aria-hidden on the already existing text. In my testing, it works exactly the same way without these changes. Can you eloborate why you added those?

src/needs-refactoring/components/CombinedFeaturePanel/components/FeatureImageUpload.tsx
```diff
@@ -153,7 +164,7 @@ export const FeatureImageUpload: FC<{
           <ImageUploadCallToAction />
         </Flex>
         <Dialog.Content>
-          <Dialog.Title>
+          <Dialog.Title ref={titleRef} tabIndex={0}>
```
kriskbx: issue (blocking):

You should never use tab-index manually. If you want to change the tab-order: Move the element in the DOM. The only reason for using tab-index is to make something static programmatically focusable with `tab-index: -1`. By using `tab-index: 0` on the headline it becomes focusable with the tab key. This is not good, only elements that you can interact with should be focusable.

src/needs-refactoring/components/CombinedFeaturePanel/components/FeatureImageUpload.tsx
```diff
@@ -136,6 +139,14 @@ export const FeatureImageUpload: FC<{
...
   }, [isDialogOpen]);
 
+  useEffect(() => {
+    if (isDialogOpen && titleRef.current) {
+      setTimeout(() => {
+        titleRef.current?.focus();
+      }, 100);
+    }
+  }, [isDialogOpen]);
+
```
kriskbx: suggestion (non-blocking):

This is a very complex and confusing implementation. The dialog has a `onOpenAutoFocus` property, which you can use to prevent the default behavior of focusing the first focusable element and instead focus the title programmatically.

src/needs-refactoring/components/CombinedFeaturePanel/CombinedFeaturePanel.tsx
```diff
@@ -27,16 +28,23 @@ export function CombinedFeaturePanel(props: Props) {
...
               isUploadDialogOpen={props.isUploadDialogOpen}
             />
           )}
 
-          {(!features || features.length === 0) && (
+          {props.isLoading && (
+            <Callout.Root mt="9">
+              <Callout.Text>{t("Loading features...")}</Callout.Text>
+            </Callout.Root>
+          )}
```
kriskbx, line 45: praise:

I really like that you added a separate loading state 👍

src/needs-refactoring/components/CombinedFeaturePanel/CombinedFeaturePanel.tsx
```diff
@@ -27,16 +28,23 @@ export function CombinedFeaturePanel(props: Props) {
   return (
     <React.StrictMode>
       <ErrorBoundary>
-        <Box>
+        <Box aria-busy={props.isLoading}>
```
kriskbx, line 31: question:

IIRC aria-busy is only useful for live-regions. What is the intention of adding it here?

src/needs-refactoring/lib/model/ac/ClientSideConfiguration.ts
```diff
@@ -53,19 +53,28 @@ export interface ClientSideConfiguration {
...
+  const translatedProductName = useTranslations(name);
+  const translatedClaim = useTranslations(claim);
+
+  if (categoryName) {
+    const translatedCategory = useTranslations(categoryName);
+    return `${translatedCategory} - ${translatedProductName}`;
+  }
+
+  if (specialisedTitle) {
+    return `${specialisedTitle} – ${translatedProductName}`;
```
kriskbx, line 26: issue (blocking):

See comments above: Move the title logic into individual pages.

src/pages/_app.tsx
```diff
@@ -93,9 +94,9 @@ const getInitialProps: typeof NextApp.getInitialProps = async (appContext) => {
   const environment = getEnvironmentVariables();
 
   // this is set by a custom middleware that handles language headers
-  const languageTag = request?.headers[
-    "x-preferred-language-tag"
-  ] as LanguageTag;
+  const languageTag =
+    (request?.headers["x-preferred-language-tag"] as LanguageTag) ||
+    fallbackLanguageTag;
```
kriskbx: question:

As far as I understand how Next middlewares work, this should be always set. Also it already falls back to the fallbackLanguageTag: https://github.com/sozialhelden/core/blob/main/lib/i18n/utils/language-tags.ts#L44

Why did you add this? Did you encounter a situation where it wasn't set?

src/pages/_document.tsx
kriskbx, line 1: suggestion (non-blocking):

I think this would have been a good opportunity to migrate from the pages router to the app router as the app router already has a language tag on the html element. I think I will give it a try and timebox it. Maybe I can do it in 2 hours. If not, let's postpone the migration.

src/utils/dispatchKeydownEvent.ts
```diff
@@ -0,0 +1,10 @@
+export const dispatchKeydownEvent = (ref: React.RefObject<HTMLElement>) => {
```
kriskbx: issue (blocking):

This is quite misleading, it doesn't dispatch a generic keydown event, but an "Enter" keydown event. I suggest renaming this.

src/needs-refactoring/components/CombinedFeaturePanel/components/FeatureHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+import AccessibleIconMarker from "~/modules/map/components/AccessibleIconMarker";
+import { useDarkMode } from "~/hooks/useTheme";
+import { useFeatureLabel } from "~/needs-refactoring/components/CombinedFeaturePanel/utils/useFeatureLabel";
+import { isWheelchairAccessible } from "~/needs-refactoring/lib/model/accessibility/isWheelchairAccessible";
+import type { AnyFeature } from "~/needs-refactoring/lib/model/geo/AnyFeature";
+
+type Props = {
+  feature: AnyFeature;
+  level: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | undefined;
+  tabIndex?: number;
```
kriskbx: issue (non-blocking):

Seems like you added `tabIndex: -1` to all usages of FeatureHeader. Maybe this is a good indication to put `tabIndex: -1` on the heading element directly and remove it from the props as this opens the gate to be used wrongly in the future.

src/needs-refactoring/components/CombinedFeaturePanel/components/FeatureHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </Button>
+        </Flex>
+
+        {placeName ? (
+          <Flex direction="column" gap="1">
+            <Heading
+              as={level}
+              size="5"
+              weight="regular"
+              ref={ref}
```
kriskbx, line 55: suggestion (non-blocking):

When I put a ref on a component, I would expect, that I had access to the root element of that component and not to an element deeply nested in it. You need the ref outside to focus the heading, so instead of doing this, I suggest using `userImperativeHandle` and expose a `focusHeading` function, so that parent components can call this function without needing to know about the implementation.

src/layouts/DefaultLayout.tsx
```diff
@@ -34,7 +34,14 @@ export function DefaultLayout({
...
         (showSidebar ? (
-          <Sidebar isExpanded={isExpanded} onIsExpandedChange={setIsExpanded}>
+          <Sidebar
+            isExpanded={isExpanded}
+            onIsExpandedChange={setIsExpanded}
+            // using assertive here even though its discouraged because we dont want the screen reader to read a
+            // list that has already changed
+            aria-live="assertive"
+            aria-relevant="additions text"
+          >
```
kriskbx: issue (blocking):

Using the list/sidebar with the screen-reader is currently not very pleasant. I think a live region is not a good solution to solve this but we should rather use focus control. E.g. move the focus into the sidebar once a filter button was clicked. And moving the focus into the next element, once you click the "more" button.

src/needs-refactoring/components/CombinedFeaturePanel/components/AccessibilitySection/EditDropdownMenu.tsx
```diff
@@ -1,24 +1,44 @@
...
+          <IconButton
+            ref={buttonRef}
+            {...triggerProps}
+            variant="soft"
+            size="2"
+            data-testid={`${tagKey}`}
+            aria-label={t("Edit description")}
+            // VoiceOver’s click command fires a plain click event but radix dropdownMenu opens the menu on keydown or
+            // pointerdown and it does not react to a click
+            onClick={() => dispatchKeydownEvent(buttonRef)}
```
kriskbx: suggestion (non-blocking):

This feels quite hacky and I'd like to propose a different solution:

```typescript
const [open, setOpen] = useState(false);

<DropdownMenu.Root open={open}>
  <DropdownMenu.Trigger 
    onPointerDown={(e) => e.preventDefault()}
    onClick={() => setOpen((prev) => !prev)}
  >
    {/* ... /*]
  </DropdownMenu.Trigger>
  {/* ... /*]
</DropdownMenu.Root>
```

src/needs-refactoring/components/CombinedFeaturePanel/components/AccessibilitySection/DescriptionEditor.tsx
```diff
@@ -5,25 +5,27 @@ import {
...
 import { PrimaryButton } from "~/components/button/PrimaryButton";
 import useSubmit from "~/needs-refactoring/components/CombinedFeaturePanel/useSubmit";
+import { supportedLanguageTagsOptions } from "@sozialhelden/core";
+import FeatureHeader from "~/needs-refactoring/components/CombinedFeaturePanel/components/FeatureHeader";
+import styled from "styled-components";
+
+const AccessibleTextArea = styled(TextArea)`
+  &::placeholder {
+    opacity: 1; 
+    aria-hidden: true;
```
kriskbx: issue (blocking):

This is not a valid CSS property, this won't do anything. There is no aria in CSS, aria is HTML only. 

I think what you're trying to achieve here, is getting the screen-reader to not read the placeholder and the label. The solution to an accessible form is to use dedicated and visible `<label>` elements for each input and removing all placeholders. Using a placeholder is IIRC considered not very good practice and should be used quite sparingly, e.g. for a search input, where an additional label element would complicate the UI.

src/needs-refactoring/components/CombinedFeaturePanel/components/AccessibilitySection/DescriptionEditor.tsx
```diff
@@ -186,13 +185,15 @@ const DescriptionEditor = ({
...
               dir="auto"
               ref={textAreaRef}
-              aria-label={t("Enter a description here")}
+              aria-label={t(
+                "Enter your description here or edit the existing one.",
+              )}
               defaultValue={addNewLanguage ? undefined : tagValue}
               value={addNewLanguage ? textAreaValue : undefined}
-              placeholder={t("Enter a description here")}
+              placeholder={t("Enter your description here")}
```
kriskbx: issue (blocking):

See above, replace with dedicated label element.

src/needs-refactoring/components/CombinedFeaturePanel/components/AccessibilitySection/EditButton.tsx
```diff
@@ -9,16 +8,20 @@ import { FeaturePanelContext } from "../../FeaturePanelContext";
 export function EditButton({
   tagKey,
   addNewLanguage,
-}: { tagKey: string; addNewLanguage: boolean }): JSX.Element {
+  ariaLabel,
+}: {
+  tagKey: string;
+  addNewLanguage: boolean;
+  ariaLabel: string;
```
kriskbx, line 15: suggestion (non-blocking):

I would prefer to pass down all the props to the icon button element instead.

src/modules/feature-panel/components/ToiletsSection.tsx
```diff
@@ -61,22 +75,41 @@ const ToiletsSection = ({ nextToilet, isLoading, tags, feature }: Props) => {
...
+            (otherTags.length === 0 && (
+              <Text>
+                {t(
+                  "There is currently no additional information for this toilet.",
+                )}
+              </Text>
+            ))}
+        </VisuallyHidden>
+        <TagList>
+          <Flex as="div" direction="row" wrap="wrap" gapX="1" gapY="2">
```
kriskbx: issue (blocking):

A `div` cannot be the child of an unordered list.

kriskbx (changes requested):
Hey, thanks a ton for your great PR. Well done 👏 
There are some things that should be addressed in my opinion. Most notably how the page title currently works and also the list/sidebar. I will take care of the latter 🙂

src/layouts/DefaultLayout.tsx
kriskbx: In order to be able to get this out more quickly, let's postpone this together with the map navigation.

kriskbx (approved):

